Issue #77, opened 2020-11-17
Trigger events command should list recently used event names in dropdown


auchenberg-stripe:
Trigger events command should list recently used event names in dropdown, so they appear on the top.

vcheung-stripe:
Here's a thought for UX: the "recently triggered" events in the dropdown should be the ones that the user triggered from clicking **Trigger new event**, rather than events created by manually typing `stripe trigger` or by other members of the account. However, there's a concern that the **Recent events** list will be confusing, since it shows _all_ the events in the account. Do you have any thoughts on this @auchenberg-stripe?

pepin-stripe:
If I may randomly chime in;
1. To me the clock icon is pretty explicit on communicating that it's recent events (but indeed can't more _more_ explicit than text, so both work for me)
2. How are events "in the account" split? Is it really for "the whole account" or can we at least filter events per api key? AFAIK we generate a CLI-specific key for each instance of the CLI, so if we can filter by key we could potentially load all events triggered by the same user (machine), and show both the ones triggered from VSCode and the ones from the CLI directly. That would be ideal because I could see users using both CLI and VSCode extension.

gracegoo-stripe:
I chatted with Pepin offline (thanks, pepin!). To summarize our discussion:
1. There does not seem to be a way to send the API key as a parameter to filter events server-side.
2. There isn't data returned in the events blob that would let us do so client-side as far as I can tell.

So I'll continue exploring Vincent's suggestion where the UX is to only show the events triggered within VSCode. There's one open question that still remains: Should the events under the "Recent Events" view be generated with the same logic? Or remain as is?

auchenberg-stripe:
Quick feedback:

1) Makes a lot of sense to align "recently used events" to those who have been triggered from within VS Code. This aligns well with how VS Code works for recently used commands, projects, and alike.

2) Recent events lists events from coming Stripe, and should list everything from the given account (how it works today). 

To me the conceptual difference between the two are source. Is the data source local or remote? Recency should align to that.

gracegoo-stripe:
Thanks for the input, Kenneth. We should be good to close this one out then!

PR #124, opened 2021-01-15
Recently used events appear on top of Trigger Events command (#77)

gracegoo-stripe:
Adding a feature that allows recently triggered events to show up on top of the quickpick view. 

As of Rev 3: Recent Events is scoped to the events triggered through the VSCode extension. This will exclude all events triggered manually by the CLI.

In the entry of the workspace state, I set a hard limit of 100 because I figured we wouldn't gain much by storing more than that, and we want a method of cleaning out the data.

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -1,8 +1,8 @@
 import * as querystring from 'querystring';
 import * as vscode from 'vscode';
 import {getExtensionInfo, showQuickPickWithItems, showQuickPickWithValues} from './utils';
+import {getRecentEvents, recordEvent} from './stripeWorkspaceState';
 import osName = require('os-name');
-import {StripeClient} from './stripeClient';
 import {StripeEventsDataProvider} from './stripeEventsView';
 import {StripeTerminal} from './stripeTerminal';
 import {StripeTreeItem} from './stripeTreeItem';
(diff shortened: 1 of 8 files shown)


src/commands.ts
```diff
@@ -208,6 +164,30 @@ export async function openTriggerEvent() {
   }
 }
 
+export async function buildTriggerEventsList(events: string[], stripeClient: StripeClient):
+    Promise<vscode.QuickPickItem[]> {
+  const historicEvents = await stripeClient.getEvents();
```
vcheung-stripe, line 169: Unfortunately I forgot to mention that the Stripe CLI only supports triggering a subset of all the possible events. Does this return any event that is untriggerable? I fear that we may have to design this differently.

gracegoo-stripe: Thanks for pointing that out, fixing!

gracegoo-stripe: Also to answer the question, yes it can:
The list events can return events that aren't in our list: https://stripe.com/docs/api/events/types

vcheung-stripe (approved):
Looks great! I tried this out locally and was able to see the 5 most recent events triggered at the top of the QuickPick, with the most recent one automatically selected. It persisted when I closed my vscode instance and reopened it. 🚢

PR #127, opened 2021-01-20
Make filename camelcase: stripeTreeViewDataProvider.ts

vcheung-stripe:
The file `StripeTreeViewDataProvider.ts` was using a different naming convention than the rest of the files. This somehow led to a build failure where other modules were referencing the camelcase version of the filename. This change renames the file to its camelcase version.


vcheung-stripe:
The GitHub build action continues to fail, but now for a different reason. Playing around with a local fork, the fix involves changing some config will be done in a separate PR: https://code.visualstudio.com/api/working-with-extensions/continuous-integration#github-actions.

PR #128, opened 2021-01-20
Update build action

vcheung-stripe:
- From the [VS Code docs](https://code.visualstudio.com/api/working-with-extensions/continuous-integration#github-actions), if we want to run VS Code in a Linux environment with GitHub Actions, we need to enable xvfb. I verified that this works on my personal fork of this repo.
- Replaced `build` with `compile` since we don't have a `build` script in our package.json.


suz-stripe (approved):
ah yes this makes sense!

Issue #133, opened 2021-01-22
Stripe CLI notification


benjamindoe:
Issue Type: <b>Bug</b>

I have installed Stripe CLI and in my VSCode settings I have defined a path to the stripe CLI. I have authenticated myself and I still get this Welcome notification to install Stripe CLI 

Extension version: 1.6.8
VS Code version: Code 1.52.1 (ea3859d4ba2f3e577a159bc91e3074c5d85c0523, 2020-12-16T16:32:10.090Z)
OS version: Linux x64 5.10.2-2-MANJARO

PR #134, opened 2021-01-22
Add tests for detecting the user has the CLI installed at the default path

vcheung-stripe:
### Notify
@stripe/developer-products 

### Summary
Add tests that verify that `StripeClient` can detect when the CLI is installed at the default path.

### Motivation
In preparation for fixing https://github.com/stripe/vscode-stripe/issues/133, which involves detecting that the CLI is installed at a custom path.

### Testing
Verify that the tests are testing the right thing by intentionally causing them to fail, then succeed.

### DIFF
--- a/src/stripeClient.ts
+++ b/src/stripeClient.ts
@@ -1,12 +1,12 @@
 'use strict';
 
-import * as fs from 'fs';
 import * as path from 'path';
 import * as vscode from 'vscode';
 import {OSType, getOSType} from './utils';
 import {Telemetry} from './telemetry';
 
 const execa = require('execa');
+const fs = require('fs');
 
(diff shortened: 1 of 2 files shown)


src/stripeClient.ts
```diff
@@ -1,12 +1,12 @@
 'use strict';
 
-import * as fs from 'fs';
 import * as path from 'path';
 import * as vscode from 'vscode';
 import {OSType, getOSType} from './utils';
 import {Telemetry} from './telemetry';
 
 const execa = require('execa');
+const fs = require('fs');
```
vcheung-stripe, line 9: For some reason, we cannot stub this module when using the `import` syntax.

gracegoo-stripe (approved):

Issue #92, opened 2020-12-11
Run CLI commands using the full path to the stripe binary instead of using the word `stripe`


vcheung-stripe:
We are running CLI commands using the `stripe` command. This could fail if a user has aliases or symlinks set up in an unorthodox or unforeseen way. We should use the user's absolute path to their stripe binary when invoking CLI commands. There is already logic in `stripeClient.ts` for finding the path.

PR #135, opened 2021-01-25
Detect that the CLI is installed at a custom path

### LINKED ISSUE #133: Stripe CLI notification

vcheung-stripe:
### Notify
cc @stripe/developer-products 

### Summary
When the extension starts up, it is not respecting the `stripe.cliInstallPath` setting to detect if the Stripe CLI is installed. This change fixes this issue. This is related to (but does not fix) #92, which requires respecting the `stripe.cliInstallPath` setting when running stripe commands in the integrated terminal.

For example, I moved `/usr/local/bin/stripe` to `/usr/local/bin/stripe2`. With this change, the extension now asks me to install when using `/usr/local/bin/stripe`, but not `/usr/local/bin/stripe2`:

Screen Shot 2021-01-25 at 3 41 18 PM
Screen Shot 2021-01-25 at 3 42 16 PM

### Motivation
Fixes #133.

### Testing
- Unit tests.
- Manually verify by moving my CLI binary, setting `stripe.cliInstallPath` to the new path, and seeing that I'm not notified to install the CLI.

### DIFF
--- a/src/stripeClient.ts
+++ b/src/stripeClient.ts
@@ -11,19 +11,17 @@ const fs = require('fs');
 const telemetry = Telemetry.getInstance();
 
 export class StripeClient {
-  isInstalled: boolean;
-  cliPath: string;
+  cliPath: string | null;
 
   constructor() {
-    this.isInstalled = false;
-    this.cliPath = '';
+    this.cliPath = null;
     vscode.workspace.onDidChangeConfiguration(this.handleDidChangeConfiguration, this);
   }
 
   private async execute(command: string) {
-    this.detectInstalled();
+    const isInstalled = await this.detectInstalled();
 
-    if (!this.isInstalled) {
+    if (!isInstalled) {
       this.promptInstall();
       return;
     }
@@ -101,7 +99,7 @@ export class StripeClient {
     }
   }
 
-  detectInstalled() {
+  async detectInstalled() {
     const defaultInstallPath = (() => {
       const osType: OSType = getOSType();
       switch (osType) {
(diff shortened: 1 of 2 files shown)


src/stripeClient.ts
```diff
@@ -168,17 +161,11 @@ export class StripeClient {
...
-      // eslint-disable-next-line no-sync
-      const isValidPath = fs.statSync(path).isFile();
-      if (isValidPath) {
-        return path;
-      }
-    } catch (err) {}
+function isFile(path: string): boolean {
+  try {
+    // eslint-disable-next-line no-sync
+    return fs.statSync(path).isFile();
```
tjfontaine-stripe, line 167: does this really need to be synchronous?

I'm not sure `.isFile` will do the right thing for you on symbolic links, you might need to `realpath` or `readlink` it, or you might just go ahead and `fs.openSync` and `fs.fstatSync` or something along those lines

src/stripeClient.ts
```diff
@@ -102,46 +102,39 @@ export class StripeClient {
...
-        // apt-get install path on ubuntu + yum install path on centOS
-        installPaths = ['/usr/local/bin/stripe'];
-        break;
-      case OSType.windows:
-        // scoop install path on Windows 10
-        const userProfile = process.env.USERPROFILE || '';
-        installPaths = [path.join(userProfile, 'scoop', 'shims', 'stripe.exe')];
-        break;
-    }
+    const defaultInstallPath = (() => {
```
gracegoo-stripe, line 105: noob question: why do we define the function and then evaluate and assign here as opposed to doing something like calling this function in line 139?

src/stripeClient.ts
vcheung-stripe: 1. It does not need to be sync.
2. TIL `realPath` can resolve symbolic links 😮 

Will take a look at making these changes.

src/stripeClient.ts
vcheung-stripe: I prefer using immediately invoked functions for assignments like this because it feels "more functional" to me, which I prefer. I feel like either approach is valid though. Do you think moving this function out would be more readable/maintainable?

gracegoo-stripe: No need, I was asking out of curiosity. Agreed that either approach works.

gracegoo-stripe (approved):

Issue #131, opened 2021-01-21
Disable telemetry when running in development


auchenberg-stripe:
We should disable telemetry when running in development to avoid polluting the usage data of the extension. Right now we may be emitting telemetry data while developing, and this shews out data.

PR #136, opened 2021-01-26
Adding ability to inject specific implementations of Telemetry into unit tests

gracegoo-stripe:
This PR converts the Telemetry class into an interface and consolidates the responsibility of creating a telemetry instance to the extensions class. This paves the way for [#131](https://github.com/stripe/vscode-stripe/issues/131). In this commit, we prevent the stripeClient unit tests from emitting data to Google Analytics by passing in a NoOp.

Commands and StripeLinter will be updated in a following PR. Those are a bit trickier to untangle.

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -3,12 +3,12 @@ import * as vscode from 'vscode';
 import {getExtensionInfo, showQuickPickWithItems, showQuickPickWithValues} from './utils';
 import {getRecentEvents, recordEvent} from './stripeWorkspaceState';
 import osName = require('os-name');
+import {GATelemetry} from './telemetry';
 import {StripeEventsDataProvider} from './stripeEventsView';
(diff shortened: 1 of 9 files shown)


vcheung-stripe (dismissed):
I like the OO principles applied here :)

I don't know if you've seen this, but one pattern I've seen that could help with choosing the right Telemetry class in runtime is to set an environment variable `NODE_ENV`. Could be useful for the other classes that depend on Telemetry.

gracegoo-stripe:
> I don't know if you've seen this, but one pattern I've seen that could help with choosing the right Telemetry class in runtime is to set an environment variable `NODE_ENV`. Could be useful for the other classes that depend on Telemetry.

Yes, environment variables seems like the right way to go!

vcheung-stripe (approved):

PR #138, opened 2021-01-27
Injecting instance of Telemetry into Commands

gracegoo-stripe:
This commit converts the Commands module into a class so that we can inject instances of telemetry and stripeTerminal. This has the benefit of making unit testing this class a bit easier, and of course allowing the caller to control which instance of telemetry to use under different circumstances. 

### Testing
I manually tested all the commands in debug mode. It should be easier to add more unit tests from now on, I started with a few for the commands class for now.

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -3,281 +3,285 @@ import * as vscode from 'vscode';
 import {getExtensionInfo, showQuickPickWithItems, showQuickPickWithValues} from './utils';
 import {getRecentEvents, recordEvent} from './stripeWorkspaceState';
 import osName = require('os-name');
-import {GATelemetry} from './telemetry';
 import {StripeEventsDataProvider} from './stripeEventsView';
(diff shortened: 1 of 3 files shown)


src/commands.ts
```diff
@@ -3,281 +3,285 @@ import * as vscode from 'vscode';
...
+      : [];
+
+    const eventsFlag = Array.isArray(options.events) && options.events.length > 0
+      ? ['--events', options.events.join(',')]
+      : [];
+
+    this.terminal.execute('listen', [...forwardToFlag, ...forwardConnectToFlag, ...eventsFlag]);
+  };
+
+  openLogsStreaming = () => {
```
gracegoo-stripe, line 127: Another note: Most of these are changing function key words so hiding whitespace changes in the review view will be helpful

vcheung-stripe (approved):
LGTM 👍

Issue #96, opened 2020-12-15
API keys that are already committed are not being linted


vcheung-stripe:
Already committed API keys are not being linted until the file becomes dirty. I suspect this could be related to #67.

vcheung-stripe:
Related: since we are relying on the git extension being installed in the user's vscode, if we remove this dependency, we can address https://github.com/stripe/vscode-stripe/issues/137 as well.

PR #140, opened 2021-01-27
Updating StripeLinter to take in an instance of telemetry and adding unit tests

gracegoo-stripe:
Note about the ReadME Update: If you have the Stripe extension turned on for all workspaces (which I think is the default), it runs the linter on this repo which will be unhappy about the test I just added.


vcheung-stripe (approved):
LGTM.

Side note: `shouldSearchFile` may have to be modified to fix #96, but I think these tests will still be valid when we fix that.

gracegoo-stripe:
> s



> Side note: `shouldSearchFile` may have to be modified to fix #96, but I think these tests will still be valid when we fix that.

Thanks, sounds good to me

PR #142, opened 2021-01-28
Disabling telemetry during in extension development mode

gracegoo-stripe:
Replaced with a local version of telemetry that emits logs to the console:
Screen Shot 2021-01-27 at 6 23 36 PM

I know the possibility of setting NODE_ENV = 'development' was mentioned but I thought it would be safer to explicitly set our own variable to use. NODE_ENV looks like a common env_var and I worry that it could be set as 'development' just because our users are using our extension while they are developing. The sessionID check is a backup in case we forget to add the special flag for whatever reason. Though, please do let know me know if there is a more robust way to do this.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -180,12 +180,12 @@ export function activate(this: any, context: ExtensionContext) {
 export function deactivate() {}
 
 /**
- * Checks for the explicit setting of the IS_DEVELOPEMENT_MODE and
+ * Checks for the explicit setting of the EXTENSION_MODE and
  * Implcitly checks by using the magic session string. This session value is used whenever an extension
  * is running on a development host. https://github.com/microsoft/vscode/issues/10272
  */
 function getTelemetry() {
-  if (process.env.IS_DEVELOPEMENT_MODE === 'true' || env.sessionId === 'someValue.sessionId') {
+  if (process.env.EXTENSION_MODE === 'development' || env.sessionId === 'someValue.sessionId') {
     console.log('Extension is running in development mode. Using local telemetry instance');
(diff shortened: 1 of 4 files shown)


gracegoo-stripe:
Addresses https://github.com/stripe/vscode-stripe/issues/131

vcheung-stripe (approved):
LGTM. I can verify the console.log messages appear when I do `npm t` and when I launch the extension development host.

vcheung-stripe (changes requested):
Sorry, I missed something on my first readthrough!

src/extension.ts
```diff
@@ -178,3 +178,18 @@ export function activate(this: any, context: ExtensionContext) {
...
 
 export function deactivate() {}
+
+/**
+ * Checks for the explicit setting of the IS_DEVELOPEMENT_MODE and
+ * Implcitly checks by using the magic session string. This session value is used whenever an extension
+ * is running on a development host. https://github.com/microsoft/vscode/issues/10272
+ */
+function getTelemetry() {
+  if (process.env.IS_DEVELOPEMENT_MODE === 'true' || env.sessionId === 'someValue.sessionId') {
```
vcheung-stripe, line 188: This checks if it's "true", do we want to check if it's "development" instead? Also nit, `IS_DEVELOPEMENT_MODE` has an extra `E` in it.

vcheung-stripe (approved):
Thanks for making the changes! 🚢

src/extension.ts
gracegoo-stripe: Good catches. Typo on my part as I was switching things around!

Issue #143, opened 2021-01-28
Survey prompt reappears even if the user has already taken the survey


vcheung-stripe:
We should resolve this so users do not get spammed with notifications.

PR #144, opened 2021-01-28
Fixing issue with SurveyPrompt not persisting data in extension context

gracegoo-stripe:
This should address https://github.com/stripe/vscode-stripe/issues/143

Changing this to an arrow function means `this` will always be in context of our SurveyPrompt class. Before, it would be in context of whoever called this showSurvey() function which is from within the `setTimeout` function, I believe. 

## Testing
1. I removed the timeout and random assignment of the survey to force the survey to show up.
2. Without this change, the we silently fail on this line `this.storage.update(StorageKeys.lastSurveyDate, currentEpoch);`
3. With this change, we persist the data. Clicking "Take survey" and then re-starting the run a second time results in no survey being prompted.

### DIFF
--- a/src/surveyPrompt.ts
+++ b/src/surveyPrompt.ts
@@ -54,7 +54,7 @@ export class SurveyPrompt {
     return true;
   }
 
-  public async showSurvey() {
+  showSurvey = async() => {
     const prompts = ['Take survey', 'Maybe later', "Don't Show Again"];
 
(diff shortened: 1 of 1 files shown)


vcheung-stripe:
LGTM. Yeah the javascript `this` doesn't work how you think it should 🚢

vcheung-stripe (approved):

PR #146, opened 2021-01-29
Run commands in vscode terminal with full path to CLI

### LINKED ISSUE #92: Run CLI commands using the full path to the stripe binary instead of using the word `stripe`

vcheung-stripe:
### Notify
cc @stripe/developer-products 

### Summary
Instead of using the string `stripe`, use the full path to the CLI executable to run Stripe commands in the vscode terminal. We already use this path when detecting if the user has the CLI installed. This also means our commands will use the user's `stripe.cliInstallPath` setting if they set a custom path.

To achieve this, we made `StripeTerminal` depend on `StripeClient` and added a public method to `StripeClient` to get the CLI path.

Screenshots:
Screen Shot 2021-01-28 at 4 10 00 PM
Clicked **Start API logs streaming**; default path was used.

Screen Shot 2021-01-28 at 4 10 50 PM
Clicked **Start API logs streaming**; custom path was used.

Screen Shot 2021-01-28 at 4 11 39 PM
Clicked **Start API logs streaming** but CLI wasn't installed; prompt for install.

Screen Shot 2021-01-28 at 4 11 12 PM
Clicked **Start API logs streaming** but CLI wasn't at custom path; show error message.

Screen Shot 2021-01-28 at 4 12 22 PM
A user can still listen for events and stream logs simultaneously.

### Motivation
Fixes #92 

### Testing
- Added tests to check that the `StripeTerminal` class will run commands with `/usr/local/bin/stripe` or a custom path
- Updated tests for `StripeClient` to reflect its new public interface
- Manually verified; see screenshots

### DIFF
--- a/src/stripeTerminal.ts
+++ b/src/stripeTerminal.ts
@@ -1,5 +1,6 @@
 import * as vscode from 'vscode';
 import {OSType, filterAsync, findAsync, getOSType} from './utils';
+import {StripeClient} from './stripeClient';
 import psList from 'ps-list';
 
 type SupportedStripeCommand = 'events' | 'listen' | 'logs' | 'login' | 'trigger';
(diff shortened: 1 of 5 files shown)


src/stripeTerminal.ts
```diff
@@ -25,17 +28,22 @@ export class StripeTerminal {
     command: SupportedStripeCommand,
     args: Array<string> = [],
   ): Promise<void> {
+    const cliPath = await this.stripeClient.getCLIPath();
+    if (!cliPath) {
```
gracegoo-stripe, line 32: I was gonna suggest we prompt the user if this happens but it looks like we already do that from getCLIPath() -> detectInstalled()!

gracegoo-stripe (approved):

PR #149, opened 2021-02-02
Create Git class to check if the user's workspace is using git

vcheung-stripe:
### Notify
@stripe/developer-products 

### Summary
To check if the user is using git, check if the workspace root contains a `.git` folder rather than checking if the vscode git extension is enabled.

We put git-related logic in a new `Git` class and have the `StripeLinter` depend on it. A little bit of refactoring in `StripeLinter` was needed to achieve this.

This does not completely remove the dependency on the vscode git extension because we are still using it to check for which files to lint, but this will be fixed in the next PR.

### Motivation
We want to know if the user is using git because we want show different error messages with our linter, but we want to do it without depending on the user having the vscode git extension.

### Testing
- Added tests for the `Git` class to verify it recognizes a workspace with and without a `.git` folder
- Manually verified that the linter messages would differ whether I had `.git` or not

### DIFF
--- a/src/stripeLinter.ts
+++ b/src/stripeLinter.ts
@@ -98,11 +98,7 @@ export class StripeLinter {
 
     const message = await this.git.isGitRepo(document.uri) ? diagnosticMessageGit : diagnosticMessageNoGit;
 
-    // get each line's possible API key warnings
-    // then flatten nested diagnostics into a flat array
-    const fileDiagnostics: Diagnostic[] = lines
-      .map(this.prepareLineDiagnostics(message))
-      .reduce((acc, next) => acc.concat([...next]), []);
+    const fileDiagnostics: Diagnostic[] = lines.flatMap(this.prepareLineDiagnostics(message));
 
     // tell VS Code to show warnings and errors in syntax
     diagnosticCollection.set(document.uri, fileDiagnostics);
--- a/src/stripeLinter.ts
+++ b/src/stripeLinter.ts
@@ -3,11 +3,13 @@ import {
   DiagnosticCollection,
   DiagnosticSeverity,
   Range,
+  TextDocument,
   extensions,
   languages,
   window,
   workspace,
 } from 'vscode';
+import {Git} from './git';
 
 import {Telemetry} from './telemetry';
 interface Resource {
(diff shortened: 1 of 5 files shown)


src/stripeLinter.ts
```diff
@@ -73,33 +75,30 @@ export const shouldSearchFile = (currentFilename: string): boolean => {
...
 
     const text = document.getText();
     const lines = text.split('\n');
 
-    // get each line's possible API key warnings
-    // then flatten nested diagnostics into a flat array
-    const fileDiagnostics: Diagnostic[] = lines
-      .map(this.prepareLineDiagnostics)
-      .reduce((acc, next) => acc.concat([...next]), []);
+    const message = await this.git.isGitRepo(document.uri) ? diagnosticMessageGit : diagnosticMessageNoGit;
```
gracegoo-stripe, line 99: Would it be reasonable to assume that this won't change over the lifecycle of the extension running?
i.e. Could we save some time by calculating this outside of StripeLinter and passing in just an `isGitRepo` flag to StripeLinter's constructor?

src/stripeLinter.ts
```diff
@@ -73,33 +75,30 @@ export const shouldSearchFile = (currentFilename: string): boolean => {
...
+
+    const fileDiagnostics: Diagnostic[] = lines.flatMap(this.prepareLineDiagnostics(message));
```
gracegoo-stripe, line 101: Is it okay to remove the reduce? Are the diagnostics not actually nested?

gracegoo-stripe (commented):
Nice! I just have a couple of questions inline.

src/stripeLinter.ts
vcheung-stripe: Good idea, I don't expect a user to be initializing git or removing git in the middle of a session very often, so we can optimize for the common case and only compute `isGitRepo` once when the extension activates.

I also plan to add more functionality to the `Git` class such as checking if a file is git-ignored (the main thing to do to remove our dependency on the git vscode extension), so I want to continue passing the `Git` instance to `StripeLinter` if that's reasonable.

src/stripeLinter.ts
vcheung-stripe: Originally we were mapping over the array, then flattening the result with `reduce`. Those two methods together should be equivalent to `flatMap`.

src/stripeLinter.ts
vcheung-stripe: Actually this could get hairy when using [multi-root workspaces](https://code.visualstudio.com/docs/editor/multi-root-workspaces). We'd have to do this per folder in the workspace. I did some benchmarking and found that `access`ing the `.git` every time adds only a few ms of latency, so I think it's okay. What are your thoughts?

src/stripeLinter.ts
gracegoo-stripe: Oh! I totally missed the `map` to `flatMap. My mistake

src/stripeLinter.ts
gracegoo-stripe: Ah good point. Yeah that seems like the best option.

gracegoo-stripe (approved):

PR #159, opened 2021-02-03
Check if file is git-ignored before linting

### LINKED ISSUE #96: API keys that are already committed are not being linted

vcheung-stripe:
### Notify
cc @stripe/developer-products 

### Summary
- Before: We depend on the user having the git vscode extension installed in order to check if a file should be linted or not.
- After: Don't use the git vscode extension to check if a file should be linted. Instead, check if the file is ignored by spawning a child process with `git check-ignore <file>`. If ignored, don't lint.

### Motivation
- Fixes #96, where API keys that are already committed are not being linted.
- Also removes a dependency on the git vscode extension.

### Testing
- Updated linter tests
- Unsure how to test the `isIgnored` method in isolation, so tested manually
- Verified manually that API keys that are already committed are linted, and git-ignored files are not linted

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "Snippets"
   ],
   "extensionKind": [
-    "ui"
+    "workspace"
   ],
   "keywords": [
     "stripe",
@@ -60,9 +60,6 @@
     "onLanguage:ruby",
     "onLanguage:java",
     "onLanguage:php"
-  ],
-  "extensionDependencies": [
-    "vscode.git"
   ],
   "main": "./out/extension.js",
   "contributes": {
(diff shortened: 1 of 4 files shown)


gracegoo-stripe (dismissed):
LGTM!

vcheung-stripe:
After discussing with @suz-stripe, we were concerned with the performance implications of spawning a process to do this. We add telemetry for the `isIgnored()` call duration to validate whether perf is a concern.

As a sanity check, I ran this locally on a very large repo and discovered it takes anywhere from 75-250ms, so it feels performant enough. But then again I have a very good laptop.

suz-stripe (approved):
👍  LGTM

PR #162, opened 2021-02-05
Add methods for spawning and ending a Stripe CLI child process (only support `stripe logs tail` for now)

### LINKED ISSUE #84: As a user I can see a list of recent logs in the activity bar view
Streamed logs should be shown in the activity bar, similar to how events are shown.

vcheung-stripe:
### Notify
cc @stripe/developer-products 

### Summary
Added two methods to `StripeClient`:
- `getOrCreateStripeProcess: (StripeProcess) => Promise<ChildProcess>`
  -  Spawns a "singleton" child process running a stripe CLI command (currently only supports `stripe logs tail`).
- `endStripeProcess: (StripeProcess) => void`
  - Ends the process if it exists

These methods will be used in `StripeLogsView` in a separate PR in order to stream logs in the UI.

### Motivation
From #84, we want to stream logs in the UI rather than in the VS Code terminal. This will be done over several PRs.

### Testing
- Added tests to verify:
  - it spawns a `stripe logs tail` process
  - it reuses the process if it already exists
  - it passes flags to the process
  - it ends the process

### DIFF
--- a/src/stripeClient.ts
+++ b/src/stripeClient.ts
@@ -9,23 +9,23 @@ import {Telemetry} from './telemetry';
 const execa = require('execa');
 const fs = require('fs');
 
-export enum StripeProcess {
+export enum StripeProcessName {
   LogsTail,
 }
 
-const stripeProcessToArgsMap: Map<StripeProcess, string[]> = new Map([
-  [StripeProcess.LogsTail, ['logs', 'tail']],
+const stripeProcessNameToArgsMap: Map<StripeProcessName, string[]> = new Map([
+  [StripeProcessName.LogsTail, ['logs', 'tail']],
 ]);
 
 export class StripeClient {
   telemetry: Telemetry;
   private cliPath: string | null;
-  private stripeProcesses: ChildProcess[];
+  private stripeProcesses: Map<StripeProcessName, ChildProcess>;
 
   constructor(telemetry: Telemetry) {
     this.telemetry = telemetry;
     this.cliPath = null;
-    this.stripeProcesses = [];
+    this.stripeProcesses = new Map<StripeProcessName, ChildProcess>();
     vscode.workspace.onDidChangeConfiguration(this.handleDidChangeConfiguration, this);
   }
 
@@ -126,11 +126,12 @@ export class StripeClient {
   }
 
   async getOrCreateStripeProcess(
-    stripeProcess: StripeProcess,
+    stripeProcessName: StripeProcessName,
     flags: string[] = [],
   ): Promise<ChildProcess | null> {
-    if (this.stripeProcesses[stripeProcess]) {
-      return this.stripeProcesses[stripeProcess];
+    const existingStripeProcess = this.stripeProcesses.get(stripeProcessName);
+    if (existingStripeProcess) {
+      return existingStripeProcess;
     }
 
     const cliPath = await this.getCLIPath();
@@ -138,7 +139,7 @@ export class StripeClient {
       return null;
     }
 
-    const commandArgs = stripeProcessToArgsMap.get(stripeProcess);
+    const commandArgs = stripeProcessNameToArgsMap.get(stripeProcessName);
     if (!commandArgs) {
       return null;
     }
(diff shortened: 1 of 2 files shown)


src/stripeClient.ts
```diff
@@ -116,6 +125,42 @@ export class StripeClient {
...
+    const commandArgs = stripeProcessNameToArgsMap.get(stripeProcessName);
+    if (!commandArgs) {
+      return null;
+    }
+
+    const projectName = vscode.workspace.getConfiguration('stripe').get('projectName', null);
+
+    const allFlags = [...(projectName ? ['--project-name', projectName] : []), ...flags];
+
+    const newStripeProcess = spawn(cliPath, [...commandArgs, ...allFlags]);
```
tjfontaine-stripe, line 151: is there a concern with ordering of flags for `stripe` vs the sub command?

src/stripeClient.ts
```diff
@@ -116,6 +125,42 @@ export class StripeClient {
...
+    this.stripeProcesses.set(stripeProcessName, newStripeProcess);
+    return newStripeProcess;
+  }
+
+  endStripeProcess(stripeProcessName: StripeProcessName): void {
```
tjfontaine-stripe, line 156: What happens if the child exits on its own, will it ever be removed from this list?

src/stripeClient.ts
vcheung-stripe: I've never seen the order matter between the global and local flags, but I'll ask around to verify.

src/stripeClient.ts
vcheung-stripe: These methods don't handle that case, but they should 👍 . Fixing.

vcheung-stripe: Added event handlers for `exit` and `error` that will remove the item from the map.

src/stripeClient.ts
vcheung-stripe: Note: Subcommand flags can't go before the subcommand itself, but global flags can go anywhere. Should be fine here.

vcheung-stripe:
I don't like having the linter/prettier changes ride along in this PR, so I am going to set this as a draft and open a new one for only the formatting changes.

src/stripeClient.ts
tjfontaine-stripe: So, I think you want to use `close` that's ultimately the event that means the process has completely gone away.

`error` is important in the event it failed to spawn the process as well (or if it failed to kill it).

I am not opposed to you listening to more than one event, but I would hoist all the cleanup logic into a single function that cleaning up. But I suspect you may have more of that in your follow up PRs when you're actually interpreting STDIO.

gracegoo-stripe (dismissed):

gracegoo-stripe:
Oops, please disregard that approve -- I meant to do so for another PR!

pepin-stripe:
Looks neat! We'll probably be doing something a bit similar for Playback, so it's good inspiration.

On nit command: I'm wondering if we should be using the naming `CLIProcess` instead of `StripeProcess` to make it a bit more explicit?

src/stripeClient.ts
vcheung-stripe: - Changed to `close` instead of `exit`.
- I moved cleanup logic into a separate function, but I don't move the `on` calls. Is that what is meant by "hoist all the cleanup logic into a single function that cleaning up?" (My goal in the end is that whichever class calls `getOrStartCLIProcess` would not have to be concerned about cleanup.)

vcheung-stripe:
> Looks neat! We'll probably be doing something a bit similar for Playback, so it's good inspiration.
> 
> On nit command: I'm wondering if we should be using the naming `CLIProcess` instead of `StripeProcess` to make it a bit more explicit?

I agree that would improve clarity. I renamed references to "stripe process" to be "CLI process" instead.

src/stripeClient.ts
tjfontaine-stripe: yes, that's what I meant by hoist encapsulating all of that cleanup logic into a common handler.

Generally this looks like directionally what we'd want, I'm sure in follow up PRs which use STDIO I'll have more feedback.

One final nit, in `endCLIProcess` you don't need to explicitly call cleanup, as either `error` or `close` should fire and cleanup would happen there. The "hard" part here is if a caller wants to kill a process and wait for it to quiesce before launching the next one, which will be tricky to handle from a promise/state machine perspective -- as is you remove the reference you're tracking and then can launch a new process (but the old process may still be out there).

tjfontaine-stripe (approved):

PR #164, opened 2021-02-06
Adding prettier config and applying to the repo

gracegoo-stripe:
Adding a project prettier [config](https://prettier.io/docs/en/configuration.html) so that we can keep our auto-formatting in sync across the developers that work on this repo.

Any suggestions on config changes are welcome. Here is the full list of options: https://prettier.io/docs/en/options.html.

I've applied the config to all files except the ones that I think are actively being worked on right now to avoid unnecessary merge conflicts. If I touched a file that you're working on, please let me know and I will omit it from this commit and we can apply these changes after you're done.

All changes to source code is purely formatting, no logic has changed.


vcheung-stripe:
Nice, thanks for setting this up. Since I'm not familiar with how prettier works, on first glance I have a couple questions:
- Do we have to do anything to keep prettier compatible with our eslint config? I found this in the docs: https://prettier.io/docs/en/install.html#eslint-and-other-linters
- How do I run prettier? Do I need to install it globally, or can we add to it to the dev dependencies or something?

gracegoo-stripe:
Good questions!

> * Do we have to do anything to keep prettier compatible with our eslint config? I found this in the docs: https://prettier.io/docs/en/install.html#eslint-and-other-linters

Short answer is no, thanks to whoever set this up previously. We already have eslint-config-prettier in our [dependencies](https://github.com/stripe/vscode-stripe/blob/162482d8468a40cff1d833906055eccad918a6b5/package.json#L307) and our eslint [config](https://github.com/stripe/vscode-stripe/blob/162482d8468a40cff1d833906055eccad918a6b5/.eslintrc.js#L6) incorporating it.

Ref: https://www.robertcooper.me/using-eslint-and-prettier-in-a-typescript-project

> * How do I run prettier? Do I need to install it globally, or can we add to it to the dev dependencies or something?
I run it via the [VSCode extension](https://prettier.io/docs/en/editors.html), you can manually format it via the command pallet or have it auto-format on save. This brings up a good point though, I think it would be nice if we added this to our [workspace recommended extensions](https://code.visualstudio.com/docs/editor/extension-gallery#_workspace-recommended-extensions). I'll update this with that

vcheung-stripe (approved):
👍  Nice!

PR #171, opened 2021-02-09
Add CLI install path on M1 Macs

### LINKED ISSUE #166: Stripe VS Code Extension can't find Stripe CLI on M1 Mac
## Issue
Running latest stable VS Code and Stripe Extension on M1 MacBook Air (see below for full version info), and the Stripe VS Code extension is unable to locate the Stripe CLI, which I can access via Terminal (see below screenshot).

Screen Shot 2021-02-06 at 9 29 15 AM

I first installed the Stripe VS Code Extension before installing the Stripe CLI. I then used Homebrew to install the Stripe CLI via Terminal. But I continued to get the same error message in the Stripe VS Code Extension, even after quitting and relaunching VS Code and uninstalling and reinstalling the  Stripe VS Code Extension. 

I wonder if the issue could be that [Homebrew uses a different path](https://docs.brew.sh/Installation) on M1 Macs (/opt/homebrew) than on Intel Macs (/usr/local). Maybe the Stripe VS Code Extension isn't checking in the right place for the CLI files on M1 Macs. 

**Update:** I think this is the case, see: 
https://github.com/stripe/vscode-stripe/blob/60504a3ab1289fcf8bffabda41ef49aa4534a962/src/stripeClient.ts#L119-L136
 
### VS Code
Version: 1.53.0
Commit: 8490d3dde47c57ba65ec40dd192d014fd2113496
Date: 2021-02-03T16:01:04.046Z
Electron: 11.2.1
Chrome: 87.0.4280.141
Node.js: 12.18.3
V8: 8.7.220.31-electron.0
OS: Darwin x64 20.3.0
### stripe.vscode-stripe
Version: 1.7.0
### OS
System Version: macOS 11.2 (20D64)
Hardware: MacBook Air (M1, 2020)

jhstrauss:
Fixes #166 which describes how the Stripe VS Code Extension cannot find the Stripe CLI on M1 Macs because the homebrew install path is different from on Intel Macs (see https://docs.brew.sh/Installation).

I don't know TypeScript, so I tried as best I could to copy existing syntax. Also, I didn't check to see if `OSType` or  `defautInstallPath` are used anywhere else that might be broken by these changes. Finally, I did not test any of my changes locally beyond a local prototype in Node.

As such, **please consider this PR more of a Proof of Concept for the fix** and confirm my changes actually work before merging :-)

### DIFF
--- a/src/stripeClient.ts
+++ b/src/stripeClient.ts
@@ -119,8 +119,11 @@ export class StripeClient {
       const osType: OSType = getOSType();
       switch (osType) {
         case OSType.macOS:
-          // HomeBrew install path on macOS
+          // HomeBrew install path on macOS Intel Macs
           return '/usr/local/bin/stripe';
+        case OSType.macOSapple:
+          // HomeBrew install path on macOS M1 Macs
+          return '/opt/homebrew/Cellar/stripe';
         case OSType.linux:
           // apt-get install path on ubuntu + yum install path on centOS
           return '/usr/local/bin/stripe';
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,7 +1,9 @@
 import * as vscode from 'vscode';
+const os = require('os'); // Not sure if this is needed/will work
 
 export enum OSType {
   macOS = 'macOS',
+  macOSapple = 'macOSapple', // Need a new OSType value for M1 Macs because Homebrew uses a different directory than on Intel Macs
   linux = 'linux',
   unknown = 'unknown',
   windows = 'windows',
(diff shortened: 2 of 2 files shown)


jhstrauss (commented):
Fixed syntax error in line 25 of utils.ts (missing `.`)

jhstrauss:
@vcheung-stripe howdy, have you had a chance to look into the linked issue re: Stripe CLI install location on M1 Macs?

tomer-stripe:
@jhstrauss sorry for the delay here! We don't have M1 machines at work yet so we haven't gotten to dig in to this. I'll try this out on my person laptop and see if I can confirm the behavior!

tomelm:
(commenting from my personal laptop)

This unfortunately isn't working for me. `os.cpus()` returns an empty array so `const cpu: string = cpus[0].model;` crashes. I'll poke around and see if I can find another option!

tomelm:
I think I found another solution! I'll open a separate PR for that. Thanks for the help here :)

jhstrauss:
@tomer-stripe thanks so much for making it actually work! :)

PR #172, opened 2021-02-09
Start and stop logs streaming in UI

vcheung-stripe:


vcheung-stripe:
@tjfontaine-stripe This is a work in progress, but do you mind taking a look to make sure we're going in the right direction?

src/stripeLogsView.ts
```diff
@@ -1,15 +1,164 @@
...
+    const stripeLogsTailProcess = await this.stripeClient.getOrCreateCLIProcess(
+      CLICommand.LogsTail,
+      ['--format', 'JSON'],
+    );
+    if (!stripeLogsTailProcess) {
+      throw new Error('Failed to start `stripe logs tail` process');
+    }
+
+    // The CLI lets you know that streaming is ready via stderr
+    if (!this.logsStderrStream) {
```
tjfontaine-stripe, line 111: If we hit this case, how do we know this is attached to the currently running process vs a previously existing process?

Generally feels like we need to effectively listen on `close` from the process to be able to trigger `cleanupStreams` logic.

src/stripeLogsView.ts
```diff
@@ -1,15 +1,164 @@
...
+      await new Promise<void>((resolve) => {
+        this.logsStderrStream = new stream.Writable({
+          write: (chunk, _, callback) => {
```
tjfontaine-stripe, line 114: probably want to `.setEncoding('utf-8')` on the stream, and then validating that the encoding here is as expected.

src/stripeLogsView.ts
```diff
@@ -1,15 +1,164 @@
...
+            const label = chunk.toString();
```
tjfontaine-stripe, line 115: Something to consider when stream processing -- you may not get an entire line in a chunk, or you may get multiple lines in the same chunk.

src/stripeLogsView.ts
```diff
@@ -1,15 +1,164 @@
...
+            if (label.includes('Ready!')) {
```
tjfontaine-stripe, line 116: checking for a status message on `stderr` makes me sad, as a personal preference I would prefer the cli emitted this to `stdout` when the format is `json`, but I guess we have to support what we've already released thus far anyway.

Something else to consider, how long should this wait to resolve before we decide something has gone wrong?

src/stripeLogsView.ts
```diff
@@ -1,15 +1,164 @@
...
+            callback();
+          },
+        });
+        stripeLogsTailProcess.stderr.pipe(this.logsStderrStream);
+      });
+    }
+
+    if (!this.logsStdoutStream) {
+      this.logsStdoutStream = new stream.Writable({
+        write: (chunk, _, callback) => {
```
tjfontaine-stripe, line 128: same notes here for encoding and the assumptions you can make about how much is in the chunk.

src/stripeLogsView.ts
```diff
@@ -1,15 +1,164 @@
...
+
+    // The CLI lets you know that streaming is ready via stderr
+    if (!this.logsStderrStream) {
+      await new Promise<void>((resolve) => {
+        this.logsStderrStream = new stream.Writable({
+          write: (chunk, _, callback) => {
+            const label = chunk.toString();
+            if (label.includes('Ready!')) {
+              resolve();
+            }
```
tjfontaine-stripe, line 118: what should we do with messages that aren't status messages?

src/stripeLogsView.ts
```diff
@@ -1,15 +1,164 @@
...
+    }
+    if (this.logsStderrStream) {
+      this.logsStderrStream.destroy();
+      this.logsStderrStream = null;
+    }
+  };
+
+  private insertLog = (logTreeItem: StripeTreeItem) => {
+    this.logTreeItems.unshift(logTreeItem);
+    this.refresh();
```
tjfontaine-stripe, line 157: does vscode debounce this for us, or do we need to be mindful of how frequently we ask the UI to redraw?

src/stripeLogsView.ts
```diff
@@ -1,15 +1,164 @@
...
+        stripeLogsTailProcess.stderr.pipe(this.logsStderrStream);
+      });
+    }
+
+    if (!this.logsStdoutStream) {
+      this.logsStdoutStream = new stream.Writable({
+        write: (chunk, _, callback) => {
+          try {
+            const logObject = JSON.parse(chunk.toString());
+            if (logObject && typeof logObject === 'object') {
```
tjfontaine-stripe, line 131: probably want to be a bit more defensive in looking for objects that smell like what we need.

tjfontaine-stripe (commented):
I like the way this is headed generally, primary feedback is to be mindful of how you process `chunk`, it may not have everything you're looking for (or in fact may have too much).

We need to also think a bit about how we want to think about the intersection of the stream handling and process lifecycle, there can be some rough edges there, where the process may restart but you're attached to the old process's streams. You could also listen to the lifecycle events from the streams (basically looking for `close` here as well) for when it's time to forget the old streams.

src/stripeLogsView.ts
```diff
@@ -1,15 +1,203 @@
...
+    const stripeLogsTailProcess = await this.stripeClient.getOrCreateCLIProcess(
+      CLICommand.LogsTail,
+      ['--format', 'JSON'],
+    );
+    if (!stripeLogsTailProcess) {
+      throw new Error('Failed to start `stripe logs tail` process');
+    }
+
+    stripeLogsTailProcess.on('close', this.cleanupStreams);
+    stripeLogsTailProcess.on('error', this.cleanupStreams);
```
tjfontaine-stripe, line 133: fwiw, you should probably only need to listen on `close` since you don't need to handle any special case for errors

src/stripeLogsView.ts
```diff
@@ -1,15 +1,203 @@
...
+
+    // The CLI lets you know that streaming is ready via stderr
+    if (!this.logsStderrStream) {
+      await new Promise<void>((resolve) => {
+        this.logsStderrStream = new stream.Writable({
+          write: (chunk, encoding, callback) => {
+            if (encoding === 'utf8') {
```
tjfontaine-stripe, line 140: now that you are `setEncoding` I'm not sure you need to be defensive here, or rather if you are I think probably want to reject the promise, or so?

src/stripeLogsView.ts
```diff
@@ -1,15 +1,203 @@
...
+      this.logsStdoutStream = null;
+    }
+    if (this.logsStderrStream) {
+      this.logsStderrStream.destroy();
+      this.logsStderrStream = null;
+    }
+    this.stripeClient.endCLIProcess(CLICommand.LogsTail);
+  };
+
+  private debouncedRefresh = debounce(this.refresh.bind(this), 1000);
```
tjfontaine-stripe, line 192: nit, prefer a constant rather than a literal

src/stripeLogsView.ts
```diff
@@ -1,15 +1,203 @@
...
+    if (!stripeLogsTailProcess) {
+      throw new Error('Failed to start `stripe logs tail` process');
+    }
+
+    stripeLogsTailProcess.on('close', this.cleanupStreams);
+    stripeLogsTailProcess.on('error', this.cleanupStreams);
+
+    // The CLI lets you know that streaming is ready via stderr
+    if (!this.logsStderrStream) {
+      await new Promise<void>((resolve) => {
```
tjfontaine-stripe, line 137: what happens if the process fails to invoke while we're waiting on this promise? Seems like we'll be notified in the `error` handler for ` stripeLogsTailProcess.on('error'` which will trigger cleaning up the streams (which haven't entirely been completed yet, but we'll never finish this?

Relatedly, what if it's taking a "long" time for the cli to report itself as being ready, will we have a deadline here?

src/stripeLogsView.ts
```diff
@@ -1,15 +1,203 @@
...
+        write: (chunk, encoding, callback) => {
+          if (encoding === 'utf8') {
+            try {
+              const object = JSON.parse(chunk);
+              if (isLogObject(object)) {
+                const label = `[${object.status}] ${object.method} ${object.url} [${object.request_id}]`;
+                const logTreeItem = new StripeTreeItem(label);
+                this.insertLog(logTreeItem);
+              }
+            } catch {}
```
tjfontaine-stripe, line 167: feels like we should be reporting this somewhere, not sure if we need to inform the customer because it's probably not actionable, unless they're using an old/broken cli/extension?

src/stripeLogsView.ts
vcheung-stripe: > what happens if the process fails to invoke while we're waiting on this promise? Seems like we'll be notified in the error handler for stripeLogsTailProcess.on('error' which will trigger cleaning up the streams (which haven't entirely been completed yet, but we'll never finish this?

If the process fails to invoke, then this promise will never resolve and the UI will stay in a loading state until the user cancels the command. So naively, I'd say this promise needs to reject when the process has an `error`. To implement that, maybe reject the promise in the handler of `stripeLogsTailProcess.on('error'`? This would mean moving this Promise instantiation one level out so that `reject` is in scope of the error handler. Would this be reasonable, or is another approach needed?

> Relatedly, what if it's taking a "long" time for the cli to report itself as being ready, will we have a deadline here?

Assuming the process doesn't error out, I can think of these scenarios:
- The CLI is getting ready, but it hasn't reported anything after a long time
  - A deadline might be appropriate here. What should we do if we reach the deadline and the CLI isn't ready? I think we should notify the user that it's taking a long time and have them decide if they want to cancel it or not. The UI can support cancelling the command while loading.
- The CLI reports some status other than "ready"
  - In this case, we should show the status to the user, stop the command and clean up.

src/stripeLogsView.ts
vcheung-stripe: We could display them in the editor, but we would have to know which messages are status messages and which aren't.

src/stripeLogsView.ts
tjfontaine-stripe: Right, all 3 cases (failed to invoke, deadline exceeded, and unexpected status message) feel like shades of the same question of what the general approach is for handling process failures. Much of that boils down to, what are we going to show the user and when? And how proactive do we need to be in taking action?

I'm not sure I have a strong opinion on how to handle the ordering of the promise and the error handling, other than asking myself does it actually matter that we need to wait to see the `Ready!` message before configuring the `stdout` handler? There's no strict dependency between these pieces, we could just blindly attach the handlers and move on and let the basic error handling take care of cleaning things up?

vcheung-stripe: - I think that's an accurate and succinct summary. I'll be coming back to this as I'm working to make sure I'm thinking about this correctly.
- That's a good point, a promise might not actually be needed here! Would simplify things if it can be removed.

src/stripeLogsView.ts
```diff
@@ -1,15 +1,211 @@
...
+
+    const stripeLogsTailProcess = await this.stripeClient.getOrCreateCLIProcess(
+      CLICommand.LogsTail,
+      ['--format', 'JSON'],
+    );
+    if (!stripeLogsTailProcess) {
+      throw new Error('Failed to start `stripe logs tail` process');
+    }
+
+    stripeLogsTailProcess.on('close', this.stopLogsStreaming);
```
tjfontaine-stripe, line 133: can `setupStreams` be called more than once? It's not a huge leak, but if it does, this will continually append the handler to the set of listeners.

tjfontaine-stripe (commented):
other than the comment about `setupStreams` being called multiple times, this looks good.

src/stripeLogsView.ts
vcheung-stripe: Reporting when the output can't be parsed would be great, but sometimes there are stdout messages that we expect to see but can't be JSON.parsed. For example, "Checking for new versions". Is it worth reporting everything that's caught here, or should we do something else?

src/stripeLogsView.ts
vcheung-stripe: Yeah `setupStreams` could be called more than once, so I added a conditional to the caller of `setupStreams`.

tjfontaine-stripe (approved):

tjfontaine-stripe (approved):

PR #176, opened 2021-02-17
Moving regex patterns to a config file

gracegoo-stripe:
Note -- pattern.ts has logic to review but is collapsed by default due to a large deletion. 

This json file will soon be updated to support multiple languages. In addition:
* Added run configurations to allow attaching to the language server for debugging purposes.
* Updated Readme for debugging the extension.
* Moved the snippets configuration into the config directory.

## Testing 
* Ran the extension and verified hover over and snippet functionality.
* And unit tests covering the server passed.


vcheung-stripe (approved):
I like it. I'm assuming the `patterns.json` will become autogenerated?

gracegoo-stripe:
> I like it. I'm assuming the `patterns.json` will become autogenerated?

Yup, that's the plan!

PR #179, opened 2021-02-25
Adding API reference config files for all langs (-dotnet)

gracegoo-stripe:
* Updated the server to lookup the pattern based on the stripe-normalized lang (i.e. javascript -> node).
* Updated the message to say 'See this method...' instead of 'See ${methodMatch}'.
  The patterns now match on opening parentheses of the method so this showed up as 'See balance.Get( in the...'.

This config already has manual overrides applied -- the PR for that in the generator is coming soon. 

Screenshots: 
Screen Shot 2021-02-25 at 10 24 34 AM

Screen Shot 2021-02-25 at 10 24 02 AM

### DIFF
--- a/config/api_ref/patterns.json
+++ b/config/api_ref/patterns.json
@@ -2,1423 +2,2993 @@
   {
     "url": "/balance/balance_retrieve",
     "regexps": {
-      "javascript": "balance\\.retrieve"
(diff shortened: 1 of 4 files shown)


vcheung-stripe (approved):
Looks great!

PR #180, opened 2021-03-01
Add commands to command palette for starting and stopping logs streaming; remove dead command

vcheung-stripe:
Before: Start logs streaming with the command palette command `stripe.openLogsStreaming`

After: Start logs streaming with `stripe.startLogsStreaming`, stop with `stripe.stopLogsStreaming`

Tested manually that the commands work.
Tested that the commands are idempotent by running `ps -A | grep "stripe logs tail"` and checking there's at most one process.

I plan to update the docs page with these commands after this is released.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,6 @@
     "onCommand:stripe.login",
     "onCommand:stripe.openCLI",
     "onCommand:stripe.openWebhooksListen",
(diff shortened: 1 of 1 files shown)


gracegoo-stripe (approved):
Nice!

PR #181, opened 2021-03-04
Refactor StripeTreeItem constructor

vcheung-stripe:
The constructor for this class is getting quite long, so we refactor it for maintainability.

Before: 
- The constructor is `constructor(label: string, commandString?: string, tooltip?: string, contextValue?: string)`.

After: 
- The constructor is `constructor(label: string, options: StripeTreeItemOptions = {})`, where `StripeTreeItemOptions` is a type that extends from `vscode.TreeItem`. This allows us (contributors to the repo) to have parameter info on hover:

Screen Shot 2021-03-04 at 11 19 41 AM

- Existing tests pass.
- Manually verified each of the tree items' functionalities continue to work.

### DIFF
--- a/src/stripeTreeItem.ts
+++ b/src/stripeTreeItem.ts
@@ -1,18 +1,25 @@
 import {TreeItem, TreeItemCollapsibleState} from 'vscode';
 
+type StripeTreeItemOptions = Pick<TreeItem, 'contextValue' | 'tooltip'> & {
+  /**
+   * The command that should be executed when the tree item is selected.
+   * Automatically prefixed with `'stripe.'`.
+   */
+  commandString?: string;
+};
+
 export class StripeTreeItem extends TreeItem {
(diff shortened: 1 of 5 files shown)


src/stripeTreeItem.ts
```diff
@@ -1,18 +1,25 @@
 import {TreeItem, TreeItemCollapsibleState} from 'vscode';
 
+type StripeTreeItemOptions = Pick<TreeItem, 'contextValue' | 'tooltip'> & {
```
vcheung-stripe, line 3: I picked only the class members from `vscode.TreeItem` we were using, then extended it with `commandString` because it's `StripeTreeItem`'s custom parameter.

gracegoo-stripe: Nice!

gracegoo-stripe (approved):
Looks great!

PR #183, opened 2021-03-09
Migrate from static resources -> ThemeIcon API

vcheung-stripe:
Use the [vscode ThemeIcon API](https://code.visualstudio.com/api/references/vscode-api#ThemeIcon) for icons instead of including our own svgs. The icons we have been using are all available through that API, so the migration is trivial.

Why migrate?
- According to the docs above: "Using a theme icon is preferred over a custom icon as it gives product theme authors the possibility to change the icons."
- Using `ThemeIcon` means adding a new icon will be trivial (when building out a stripe samples UI, for example).

Before:
Screen Shot 2021-03-09 at 10 56 29 AM

After:
- The difference appears to be that the terminal icon is slightly different.
Screen Shot 2021-03-09 at 10 47 15 AM

After (light theme):
Screen Shot 2021-03-09 at 10 47 02 AM

Tested manually.


gracegoo-stripe (approved):
Looks awesome!

PR #189, opened 2021-03-12
Abstracting out streaming components of LogsView

gracegoo-stripe:
Moving out streaming management components from the stripeLogView file to an abstract class called StreamingViewDataProvider. The goal was to move the stream and state management to the abstract class, but not so strictly that we wouldn't be able to support future usecases.

This is a precursor to adding events streaming.

## Testing
Manually tested the log streaming functionality.

### DIFF
--- a/src/stripeStreamingView.ts
+++ b/src/stripeStreamingView.ts
@@ -36,7 +36,7 @@ export abstract class StreamingViewDataProvider extends StripeTreeViewDataProvid
     this.viewState = ViewState.Idle;
   }
 
-  starStreaming = async () => {
+  startStreaming = async () => {
     if (this.viewState === ViewState.Idle) {
       this.setViewState(ViewState.Loading);
       try {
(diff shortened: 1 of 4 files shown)


src/stripeStreamingView.ts
```diff
@@ -0,0 +1,175 @@
...
+  abstract createStreamProcess(): Promise<ChildProcess>;
+
+  // Tell us how we can tell when the process is ready
+  abstract streamReady(chunk: any): boolean;
+
+  // Tell us how we can tell if the stream is still loading
+  abstract streamLoading(chunk: any): boolean;
+
+  // Tell us how to process the chunk into a tree item.
+  abstract createTreeItem(chunk: any): StripeTreeItem | null;
```
vcheung-stripe, line 150: Can we make this name more specific? For example, `createStreamTreeItem` or something to that effect.

vcheung-stripe (approved):
💯  Had a non-blocking comment about a function name. Looks great!

src/stripeStreamingView.ts
gracegoo-stripe: absolutely! Will update

vcheung-stripe (approved):

vcheung-stripe (approved):

PR #190, opened 2021-03-15
Refactor activate function

vcheung-stripe:
The `activate` function is getting long, and some of the variable names are inconsistent, which adds friction when we want to add new features. This PR renames some variables for consistency, removes some redundant comments, and refactors how commands are registered.

Verified manually and with existing tests.


gracegoo-stripe (approved):
Looks great, thanks for the clean-up!

PR #192, opened 2021-03-17
Upgrading to 1.7.1

gracegoo-stripe:
Summary of changes since 1.7.0:
* Removed dependency on the VSCode Git extension
* Request log streaming in the Logs View UI
* Adding API reference support for all langs except .NET
* Adding snippet support for all langs
* Localized timestamps
* Migrated to use ThemeIcon from VSCode


vcheung-stripe (approved):
LGTM

PR #197, opened 2021-03-19
Add webpack

### LINKED ISSUE #193: Bundle Extension
When packaging the extension we get this suggestion from VSCode:
```
This extension consists of 1530 files, out of which 847 are JavaScript files. For performance reasons, you should bundle your extension: https://aka.ms/vscode-bundle-extension . You should also exclude unnecessary files by adding them to your .vscodeignore: https://aka.ms/vscode-vscodeignore
```

vcheung-stripe:
### Summary

Fixes #193.

Add webpack following the guide [here](https://code.visualstudio.com/api/working-with-extensions/bundling-extension), with modifications because our project has a different structure than what's in the guide.

Webpack will be used for prod builds and dev builds instead of tsc; tsc will continue to be used for tests. Webpack builds will live in `dist`; tsc in `out`.

What we did:
- Installed dependencies for webpack and typescript compatibility with webpack
  - This involved bumping the typescript version from 3 to 4
- Added webpack configs for the extension and the language server.
  - These are actually separate modules, so we need to webpack each one separately. This wasn't necessary before because tsc would just build everything.
- Added npm scripts for running webpack in dev mode and prod mode
  - Dev mode for faster compilation, prod mode for smaller bundle
- Updated `package.json` so that `vsce package` will use the webpack prod build instead of the tsc build
- Updated the `launch.json` to use the webpack dev build instead of the tsc build for running the extension locally
  - This involved updating `tasks.json` to use the new npm scripts we made
  - This makes the build step a little longer when developing locally because tsc was faster. This is a tradeoff for making the dev build closer to the prod build.
- Move the tests folder from src directory -> root directory
  - Tests would not work otherwise; unsure why
- Updated `tsconfig.json`'s `rootDirs`
  - In order for files outside of `src` to be importable, like the hover methods patterns.
  - In order to build the tests, which now live in the root directory instead of src.

Comparing the bundle size
- Before: 3.56 MB
- After: 2.32 MB
- Difference: 1.24 MB or 35 percent!

### Testing
- `npm test` works (uses tsc build with one exception, which uses the webpack dev build. not sure how to get around this.)
- launching and relaunching the extension development host works (uses webpack dev build)
- `vsce publish` works (uses webpack prod build)


src/stripeLanguageServer/patterns.ts
```diff
@@ -1,5 +1,6 @@
 /* eslint-disable no-warning-comments */
-import * as path from 'path';
+import patterns from '../../config/api_ref/patterns.json';
```
vcheung-stripe, line 2: This file would otherwise not be included in the webpack build.

vcheung-stripe:
Having some issues with the Github build action - setting this to a draft temporarily.

.gitignore
```diff
@@ -3,4 +3,5 @@ node_modules
 .vscode-test/
 *.vsix
 .DS_Store
-src/test/workspace/.vscode
+test/workspace/.vscode
+dist
```
gracegoo-stripe, line 7: nit: PR tool detected there's no new line at the end of this file. 

I don't know that this will cause any issues in this case but I have seen some tools not parse the files correctly because of it before.

.vscodeignore
```diff
@@ -1 +1,7 @@
-**/*.ts
\ No newline at end of file
+.vscode
+node_modules
```
gracegoo-stripe, line 2: yay!

gracegoo-stripe (approved):
Looks great! Just a non-blocking nit.

vcheung-stripe:
ptal @gracegoo-stripe. Rebased on master and verified `npm test` and launching the extension host still work.

gracegoo-stripe (approved):

Issue #184, opened 2021-03-09
Define Min Version of Stripe CLI 


gracegoo-stripe:
We should have the ability to define the minimum version of the Stripe CLI our extension is compatible with and notify the user if their CLI is installed but out of date.

PR #199, opened 2021-03-30
Adding min cli version detection

gracegoo-stripe:
Adding a min version check in stripeClient. The current recommended minimum version will be v1.5.12 because that has a required flag we want to use for the event streaming. 

This will prompt the user in bottom right, similar to the notifications we get when other extensions are out of date. It's not as intrusive as when the user does not have the CLI installed. I also added this check to the detectInstallation function because I figure we'd want to run this every time we use the CLI, similar to the installation notification.

## Testing
When updating the min version to a non-existing v.1.13: 
Screen Shot 2021-03-30 at 11 46 09 AM

This addresses https://github.com/stripe/vscode-stripe/issues/184

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,9 +10,11 @@
       "dependencies": {
         "@types/universal-analytics": "^0.4.4",
         "byline": "^5.0.0",
+        "compare-versions": "^3.6.0",
         "execa": "^4.0.0",
         "moment": "^2.24.0",
         "os-name": "^3.1.0",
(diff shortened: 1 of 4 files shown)


vcheung-stripe (dismissed):
LGTM 👍

gracegoo-stripe:
Rebased and updated the module path for the test.
ptal @vcheung-stripe

vcheung-stripe (approved):
LGTM 👍

PR #203, opened 2021-04-02
Add clear recent events functionality and replacing event details tex…

gracegoo-stripe:
…t document source

* The event details document was retrieved from the CLI with the eventId. Since we added event streaming we now have access to the entire blob. This change updates the event streaming to store this data in workspace context and updates the text document provider to fetch from there as well.
* Added a clear recent events functionality. This also clears the event data from the workspace context.

## Note for the reviewer
You may be wondering why I didn't just keep the event data in the TreeItem itself. Originally, that's what I did and bypassed the `stripeEventDocumentationProvider` entirely by just creating a document within the `openEventDetails` command. However, if you create a document with the content, it gets saved to the "Untitled" schema, which not only is an unhelpful name, but is also editable. 

A custom textDocumentProvider allows us to create readonly documents to the editor, but the interface doesn't allow us to pass in the content when we invoke the provider, only the URI. So I opted for keeping the text document provider, but replacing the stripeClient with extensionContext as the provider source. 

## Testing
There is no visible difference for the open event details UI, but I manually tested it to make sure that it's still working as it did before.

Also manually tested the clear events functionality, and added some unit tests for the management of the new storage key.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
     "onCommand:stripe.startLogsStreaming",
     "onCommand:stripe.stopLogsStreaming",
     "onCommand:stripe.clearRecentLogs",
+    "onCommand:stripe.clearRecentEvents",
     "onCommand:stripe.startEventsStreaming",
     "onCommand:stripe.stopEventsStreaming",
     "onLanguage:typescript",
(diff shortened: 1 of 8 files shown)


vcheung-stripe (dismissed):
Great! One thing we can do now that the document opens instantly is we can get rid of the progress indicator: https://github.com/stripe/vscode-stripe/blob/f4e7bed905b8b3b4d917dc743064a2a3a2e7eaa0/src/commands.ts#L222-L232. This is minor, so approving.

gracegoo-stripe:
> Great! One thing we can do now that the document opens instantly is we can get rid of the progress indicator:

Ooh, good call. Updated!

vcheung-stripe (approved):
LGTM

PR #206, opened 2021-04-13
Deleting unused code and updating cliPath initialization

gracegoo-stripe:
Previously, we had to know to call `getCLIPath` instead of using `this.cliPath` because the former actually kicked off the function that populates `this.cliPath`. This commit updates the stripeClient object to kick off the installation detection in the constructor and changes the cliPath to be a promise so that the member variable will always be updated at least once.

The getCLIPath still exists as a wrapper function clients can call. It will re-check the installation path before vending it in case it is deleted within the lifespan of stripeClient. I also updated this function to verify the version and authentication before vending it as well.

In addition:
* Adding extensionContext to StripeClient for future use with telemetry.
* Removing some dead code

## Testing
Updated unit tests and manually verified that:
* streaming still works as expected
* terminal command (trigger) works as expected
* If I move the path to the stripe executable, we get installation prompt when we try to execute a command (trigger). 
* If I update my stripe cliPath setting to an invalid one, I get custom path error.
* If I log out and try to execute a command after I get a login prompt.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -27,7 +27,7 @@ export function activate(this: any, context: ExtensionContext) {
   const telemetry = getTelemetry();
   telemetry.sendEvent('activate');
 
-  const stripeClient = new StripeClient(telemetry);
+  const stripeClient = new StripeClient(telemetry, context);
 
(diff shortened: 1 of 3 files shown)


vcheung-stripe (approved):
LGTM.

PR #208, opened 2021-04-14
Add new telemetry implementation that sends events to r.stripe.com

gracegoo-stripe:
* To send additional dimensions, updated stripeClient to set accountId and cliVersion in the workspaceState
* Created a temporary telemetry instance that will write to both sources
* Removed ip as a dimension

## Testing 
Verified events are sent to r.stripe.com via server side logs. Added some unit testing around version check and authentication


package.json
```diff
@@ -409,8 +411,10 @@
     "os-name": "^3.1.0",
     "proxyquire": "^2.1.3",
     "ps-list": "^7.1.0",
-    "public-ip": "^4.0.1",
+    "query-string": "^7.0.0",
```
vcheung-stripe, line 414: Nit: is this dependency needed? Node has a builtin `querystring` module: https://nodejs.org/api/querystring.html#querystring_querystring_stringify_obj_sep_eq_options

vcheung-stripe (approved):
This looks great!

package.json
gracegoo-stripe: Oh thank you! Removed the dependency.

src/telemetry.ts
```diff
@@ -58,7 +72,7 @@ export class GATelemetry implements Telemetry {
     return this._isTelemetryEnabled;
   }
 
-  async setup() {
+  setup() {
     if (!this.isTelemetryEnabled) {
```
vcheung-stripe, line 76: I think this needs parens:

```suggestion
    if (!this.isTelemetryEnabled()) {
```

gracegoo-stripe: Ah, I thought already added that but apparently I did not! Good catch, updated!

vcheung-stripe (approved):
LGTM

PR #209, opened 2021-04-15
Land windows fix

vcheung-stripe:
### What was broken?
Certain commands were not working on Windows, specifically the commands that involve sending text to the vscode integrated terminal: Trigger new event, Start webhooks listening*, etc. There was no problem on Mac or Linux.

### Why weren't they working on only Windows?
Some context: some of these commands are long running (e.g. Start webhooks listening) and some are short running (e.g. Trigger new event). We initially wanted those short running commands to reuse one terminal rather than splitting off a new one every time. Imagine if you wanted to trigger `checkout.session.completed` twenty times, you wouldn't want twenty terminals to appear.

It turned out there were vscode api limitations with the integrated terminal. We couldn't easily figure out what process was running on which terminal, which meant the extension couldn't easily figure out which terminal to reuse. We used the `ps-list` node module (https://www.npmjs.com/package/ps-list) for a somewhat hacky solution to that.

The `ps-list` node module was causing issues for windows after we started compiling with webpack. The module depends on a precompiled binary which wasn't being included in the build. Because of this, commands that depended on `ps-list` would not work.

### Solution (this PR)
We decided to remove the module rather than try to adapt it for windows as it has been a source of importability for a while. It already has limited windows support such that it doesn't even accomplish what we wanted initially, which was reusing terminals for short running commands.

So if we're not reusing terminals for short running commands, does that mean if a user runs "Trigger new event" twenty times, they'll get twenty terminals popping up? No because we are now sending the output of "Trigger new event" to an OutputChannel:
Screen Shot 2021-04-14 at 6 50 56 PM

This means users won't be able to see the outputs of "Start webhooks listening" and "Trigger new event" side by side. This is the tradeoff we're taking by doing this.

As an additional remediation, the next PR will be a github action to build and test on various OSes, which should hopefully be enough to catch any regressions.

### Testing
- updated tests
- ran `npm test` on Mac and Windows
- tested manually on Mac and Windows

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,6 @@
         "moment": "^2.24.0",
         "os-name": "^3.1.0",
         "proxyquire": "^2.1.3",
-        "ps-list": "^7.1.0",
         "toml": "^3.0.0",
         "universal-analytics": "^0.4.20",
         "uuid": "^3.4.0",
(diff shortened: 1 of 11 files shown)


gracegoo-stripe (approved):
This looks awesome!

PR #210, opened 2021-04-15
Update GH action to build and test on multiple OSes

vcheung-stripe:
We were running this on ubuntu-latest only. Updated to run on ubuntu-latest, macos-latest, and windows-latest. Also updated to run on pushes and PRs to all branches instead of only master.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,23 +2,33 @@ name: Build
 
 on:
   push:
-    branches: [ master ]
+    branches: ['**']
   pull_request:
-    branches: [ master ]
+    branches: [master]
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+    runs-on: ${{ matrix.os }}
+
     steps:
-    - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -1,24 +1,30 @@
...
+          node-version: ${{ env.NODE_VERSION }}
 
-    - name: Use Node.js ${{ env.NODE_VERSION }}
-      uses: actions/setup-node@v1
-      with:
-        node-version: ${{ env.NODE_VERSION }}
+      - name: npm install, and test (ubuntu)
+        if: ${{ matrix.os == 'ubuntu-latest' }}
+        run: |
+          npm install
```
suz-stripe, line 23: extremely minor nit but we could dry this up by making `npm install` its own step before branching out into the different OS steps for the test command

suz-stripe (approved):
this LGTM as is! 👍 
I'll leave it up to you as to whether you wanted to roll my suggestion into this PR as it's super minor

.github/workflows/build.yml
vcheung-stripe: Good suggestion, I'll do that.

gracegoo-stripe (approved):
Thank you so much for adding this!

PR #212, opened 2021-04-15
Adding a survey version to prompt check

gracegoo-stripe:
I've noticed a few survey responses coming in where the feedback is something like "I've already taken this survey". Adding in a concept of survey versioning so that we don't spam folks who've already taken the survey when their 12 weeks expire.

We keep the 12 week limit in there so that we don't spam the same user again within that period of time even if we release a new version of the Survey.

Users are still able to open the survey and leave feedback any time they want by clicking "Rate and provide feedback".

### DIFF
--- a/test/suite/surveyPrompt.test.ts
+++ b/test/suite/surveyPrompt.test.ts
@@ -1,0 +1,101 @@
+import * as assert from 'assert';
+import * as vscode from 'vscode';
+import {SURVEY_VERSION, StorageKeys, SurveyPrompt} from '../../src/surveyPrompt';
+import {TestMemento, mocks} from '../mocks/vscode';
+import sinon from 'sinon';
+
+suite('surveyPrompt', () => {
+  let sandbox: sinon.SinonSandbox;
+  let extensionContext: vscode.ExtensionContext;
+  let globalState: vscode.Memento;
+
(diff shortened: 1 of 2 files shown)


test/suite/surveyPrompt.test.ts
```diff
@@ -0,0 +1,101 @@
...
+    });
+  });
+
+  suite('tookMostRecentVersionOfSurvey', () => {
+    test('returns false if survey was never taken', () => {
+      const surveyPrompt = new SurveyPrompt(extensionContext);
+      assert.strictEqual(surveyPrompt.tookMostRecentVersionOfSurvey(), false);
+    });
+
+    test('returns true if older version of survey was taken', () => {
```
vcheung-stripe, line 82: Nit: I think this needs to be
```suggestion
    test('returns false if older version of survey was taken', () => {
```

vcheung-stripe (dismissed):
Thanks for the tests, it makes me confident this is correct 👍

test/suite/surveyPrompt.test.ts
gracegoo-stripe: good catch, updated!

vcheung-stripe (approved):

PR #214, opened 2021-04-19
Remove LocalTelemetry logging for development mode

gracegoo-stripe:
It spams logs in development and in unit tests. When we are debugging telemetry we can always swap out this NoOp with the logging implementation.


vcheung-stripe (approved):
Makes sense not to log these 👍

PR #223, opened 2021-05-11
Upgrading packages to address dependabot alerts

gracegoo-stripe:
The [trim](https://github.com/stripe/vscode-stripe/security/dependabot/package-lock.json/trim/open) dependency required a major version upgrade to remark, which is used to generate our 3P notices. I followed their migration guide here to update our script: https://github.com/remarkjs/remark/releases/tag/13.0.0

## Testing
ran the `node scripts/generateThirdPartyList.js ` script and added the output to the README.


vcheung-stripe (approved):
Looks like my commit was added on top, but I think it should be fine to merge.

gracegoo-stripe:
> Looks like my commit was added on top, but I think it should be fine to merge.

Thank you! Holding off on this, because I found one more issue while trying to upgrade our other (non-offending) packages

PR #224, opened 2021-05-11
Use a welcome view instead of modal to prompt CLI installation

### LINKED ISSUE #57: Use welcome views for login and install screens
We should use the new welcome views to show install and login states instead of notfications.

image

image

Feature:https://code.visualstudio.com/api/extension-guides/tree-view#welcome-content
API: https://github.com/microsoft/vscode/issues/89080

vcheung-stripe:
fixes #57 

When the CLI can't be detected, show a welcome view on startup:
Screen Shot 2021-05-11 at 12 13 15 PM

After CLI is installed, refresh to see the logs and events views:

https://user-images.githubusercontent.com/71457708/117872968-21dc1300-b254-11eb-840e-ddb01a0af05b.mov

Authentication is not a part of this flow. We may want to improve this later:

https://user-images.githubusercontent.com/71457708/117872967-21437c80-b254-11eb-93d1-70b3938abcdb.mov

If the CLI is uninstalled while the extension is active, commands will fail and the welcome view will be focused:

https://user-images.githubusercontent.com/71457708/117872961-20aae600-b254-11eb-8909-51cdb269614d.mov

Same scenario, but the stripe extension is not in view. The welcome view will be focused:

https://user-images.githubusercontent.com/71457708/117872955-1e488c00-b254-11eb-9bf9-3cd8691d9b58.mov

### DIFF
--- a/package.json
+++ b/package.json
@@ -211,12 +211,19 @@
     "views": {
       "stripe": [
         {
+          "id": "stripeInstallCLIView",
+          "name": "Install Stripe CLI",
+          "when": "stripe.isCLIInstalled == false"
+        },
+        {
           "id": "stripeEventsView",
-          "name": "Events"
+          "name": "Events",
+          "when": "stripe.isCLIInstalled == true"
         },
         {
           "id": "stripeLogsView",
-          "name": "Logs"
+          "name": "Logs",
+          "when": "stripe.isCLIInstalled == true"
         },
         {
           "id": "stripeDashboardView",
@@ -228,6 +235,12 @@
         }
       ]
     },
+    "viewsWelcome": [
+      {
+        "view": "stripeInstallCLIView",
+        "contents": "Welcome! The Stripe VS Code extension requires the Stripe CLI.\n[Install Stripe CLI](https://stripe.com/docs/stripe-cli#install)\n[Set a custom path to Stripe CLI](command:workbench.action.openSettings?%5B%22stripe.cliInstallPath%22%5D)\nLearn more about [how to use the Stripe extension for Visual Studio Code](https://stripe.com/docs/stripe-vscode)."
+      }
+    ],
     "menus": {
       "view/title": [
         {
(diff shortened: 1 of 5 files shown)


src/stripeTerminal.ts
```diff
@@ -22,6 +22,13 @@ export class StripeTerminal {
       return;
     }
 
+    if (command !== 'login') {
```
gracegoo-stripe, line 25: Good catch, it makes sense to move it auth logic out of the getCLIPath

package.json
```diff
@@ -228,6 +235,12 @@
         }
       ]
     },
+    "viewsWelcome": [
+      {
+        "view": "stripeInstallCLIView",
+        "contents": "Welcome! The Stripe VS Code extension requires the Stripe CLI.\n[Install Stripe CLI](https://stripe.com/docs/stripe-cli#install)\n[Set a custom path to Stripe CLI](command:workbench.action.openSettings?%5B%22stripe.cliInstallPath%22%5D)\nLearn more about [how to use the Stripe extension for Visual Studio Code](https://stripe.com/docs/stripe-vscode)."
```
gracegoo-stripe, line 241: In case you missed it, Tomer had some comments in the vscode channel about making this message a bit softer with regards to it only being required for many features (as opposed to all).

test/suite/telemetry.test.ts
```diff
@@ -21,19 +21,19 @@ suite('GATelemetry', function () {
         workspaceFolder,
       );
 
-      await telemetryConfig.update('enableTelemetry', false);
+      await telemetryConfig.update('enableTelemetry', false, true);
```
gracegoo-stripe, line 24: I don't think I follow why we need this here, could you explain it for me?

gracegoo-stripe (commented):
Nice! Just have some questions regarding the test updates

gracegoo-stripe (approved):

PR #225, opened 2021-05-11
Fix tests that use workspace

vcheung-stripe:
- Some telemetry tests were failing because of timeout. They make several vscode API calls, which not only slows down the tests, but also makes the test output very noisy. This PR replaces those calls with stubs.
- Those vscode API calls relied on a dummy workspace. Since we don't make those calls anymore, I removed the dummy workspace folder.
- I also updated a linter test that was failing after removing the dummy workspace folder.


gracegoo-stripe (approved):
🎉 

Thanks so much for fixing these!

PR #228, opened 2021-05-17
Removing Google Analytics Telemetry

gracegoo-stripe:
Transition to using the in-house analytics service has been live for a few weeks now, removing Google Analytics as a telemetry destination.

### DIFF
--- a/test/suite/telemetry.test.ts
+++ b/test/suite/telemetry.test.ts
@@ -1,7 +1,7 @@
 import * as assert from 'assert';
 
 import * as vscode from 'vscode';
-import {GATelemetry, StripeAnalyticsServiceTelemetry} from '../../src/telemetry';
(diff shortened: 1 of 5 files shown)


test/suite/telemetry.test.ts
```diff
@@ -18,32 +18,6 @@ suite('GATelemetry', function () {
   teardown(() => {
     sandbox.restore();
   });
-
```
vcheung-stripe, line 21: Looks like we could remove the entire suite starting from line 10

vcheung-stripe (dismissed):
I only have a nonblocking nit, otherwise LGTM.

test/suite/telemetry.test.ts
gracegoo-stripe: You're right! Removing

vcheung-stripe (approved):

PR #229, opened 2021-05-17
Adding linting to the GH action

vcheung-stripe:
This runs the linter before each test.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,6 +20,9 @@ jobs:
       - name: Install dependencies
         run: npm install
 
+      - name: Lint
+        run: npm run lint
+
       - name: Run tests (ubuntu)
         if: ${{ matrix.os == 'ubuntu-latest' }}
         run: xvfb-run -a npm run test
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -10,6 +10,12 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     steps:
+      # Windows throws false positives with linting because of CRLF
+      - name: Set git to use LF
+        run: |
+          git config --global core.autocrlf false
```
gracegoo-stripe, line 16: For my understanding, not setting this would cause the `npm run lint` to fail below?

I'm also a bit confused on the right settings -- the [docs](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration) say that setting this to false will allow CRLF endings? Is that what we want for all platforms?

vcheung-stripe: Yes, the linting fails without `git config --global core.autocrlf false`, example: https://github.com/vcheung-stripe/vscode-stripe/runs/2603830207?check_suite_focus=true

My understanding is that git automatically converts LF to CRLF, and setting `core.autocrlf false` turns off the conversion.

gracegoo-stripe: I see, thanks for explaining!

gracegoo-stripe (approved):

PR #232, opened 2021-05-17
Update events buttons

### LINKED ISSUE #70: Menu items for commands should reflect currently running commands
When starting long running commands like `stripe listen` or `stripe logs tail` menu items in the left hand panel should update accordingly, so they don't say "start X" but instead says "Stop x".

User feedback
> Can you toggle the text in the activity window? It always says “Start” after I start but people shouldn’t be able to start multiple

> Would expect to be able to “Stop webhooks listening” from the “Events” menu, rather than having to ^C in the terminal

### LINKED ISSUE #113: Make it more clear what the forward webhooks event dialog does
When starting webhooks listening we show a “Do you want to forward webhook events” dialog without much explanation. We should add a descriptive text of the functionality in the prompt.

If might not be possible due to limitations of VS Code UI.

vcheung-stripe:
- Changed command name: **Start webhooks listening** ->  **Forward events to your local machine** (closes #70)
- Added a tooltip on hover: "Forward webhook events from Stripe to your local machine by connecting directly to Stripe's API." (fixes #113)
- Because we changed the name of the button to be "Forward events ...", remove the prompt that asks if you want to forward events (because obviously you do)
- Lowercased for consistency: **Start streaming Events** -> **Start streaming events**

Before:
Screen Shot 2021-05-17 at 4 20 45 PM

After:
Screen Shot 2021-05-17 at 4 20 28 PM

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -9,7 +9,7 @@ import {
   setConnectWebhookEndpoint,
   setWebhookEndpoint,
 } from './stripeWorkspaceState';
-import {getExtensionInfo, showQuickPickWithItems, showQuickPickWithValues} from './utils';
+import {getExtensionInfo, showQuickPickWithItems} from './utils';
 import osName = require('os-name');
 import {StripeEventsViewProvider} from './stripeEventsView';
 import {StripeLogsViewProvider} from './stripeLogsView';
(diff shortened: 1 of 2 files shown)


gracegoo-stripe (dismissed):
LGTM!

vcheung-stripe:
"Forward webhook events from Stripe's API to your local machine."

vcheung-stripe:
Updated the hover text:
Screen Shot 2021-05-18 at 9 53 57 AM

gracegoo-stripe (approved):

PR #234, opened 2021-05-25
Update link to upgrade CLI

### LINKED ISSUE #216: "Update CLI" alert message needs to be updated
If user's CLI is outdated, we suggest that they update.

"We recommend being on at least v1.5.13 of the CLI for the best experience."

However, today's update link takes the user to the installation section of the documentation:
https://stripe.com/docs/stripe-cli#install

Suggest we either:

- Direct user to CLI documentation section that describes how to update CLI (which doesn't exist today, we'll have to write it).  
- Add a `stripe update` command to CLI, and reference that in both VS Code, as well as the CLI error message: "A newer version of the Stripe CLI is available, please update to: v1.5.14 by running `stripe update`"

cc: @pepin-stripe @tomer-stripe

vcheung-stripe:
fixes #216

Will merge once CLI upgrading instructions are available at https://stripe.com/docs/stripe-cli#upgrade.

### DIFF
--- a/src/stripeClient.ts
+++ b/src/stripeClient.ts
@@ -93,7 +93,7 @@ export class StripeClient {
       ...[actionText],
     );
     if (returnValue === actionText) {
-      vscode.env.openExternal(vscode.Uri.parse('https://stripe.com/docs/stripe-cli#install'));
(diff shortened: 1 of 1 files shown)


gracegoo-stripe (approved):
Looks great, thanks for updating this!

PR #236, opened 2021-05-26
Move dotnet server into vscode repo and connect language client

gracegoo-stripe:
This PR introduces the dotnet language server and connects the language client to it. This is not a complete implementation, below is a checklist of things left TODO. This will also merge into a dotnet-server branch until everything is done.

## APIReferenceHandler
This is the main logic that handles what content to return when the user hovers over a document we are interested in. We are using the semantic model to get the fully name-spaced type of the method invocation. Previously, I had tried to do this with the syntax tree which also worked but we had to handle all the different syntax for a method invocation. For example, we had to handle each of these cases ourselves: `priceService.GetAsync()`, `new PriceService().GetAsync()`, etc. The semantic model is the compiled version of the code so we can simply extract the details we need from the symbol once we find it. With this method, we are blindly trying to get the type and name of the symbol, even if it may not be a method invocation. The key is then used to lookup the api reference path in the configuration file and returns a null Hover instance if it cannot be found, and a real hover instance if it is found. 

Demo: 
https://user-images.githubusercontent.com/75757829/119735293-f569ee80-be30-11eb-9612-651c80889412.mov

## Checklist
- [x] Set up dotnet server and connect it to the extension
- [x] Implement API reference handler with semantic model analysis. 
- [ ] Implement Text document hander (I had previously done this with a buffer manager, but I want to see if I can add changes directly to the MSBuildWorkspace instance.
- [ ] Unit Testing
- [ ] Channel logs to the output channel of the extension -- right now only errors are getting through.
- [ ] Client-side: Add conditional server startup logic based on workspace project 
- [ ] Client-side: Update dotnet executable fetching -- right now it's hard-coded 
- [ ] Get the project file from the client
- [ ] Update debugging steps since we've merged the repos
- [ ] Update the configuration file with real entries
- [ ] Webpack building?


src/stripeLanguageServer/client.ts
```diff
@@ -1,12 +1,26 @@
...
 
 export class StripeLanguageClient {
-  static activate(context: ExtensionContext, serverOptions: ServerOptions, telemetry: Telemetry) {
-    const clientOptions: LanguageClientOptions = {
+  static activate(
+    context: vscode.ExtensionContext,
+    serverOptions: ServerOptions,
+    telemetry: Telemetry,
+  ) {
+    const outputChannel = vscode.window.createOutputChannel('Stripe Lanaguage Client');
```
vcheung-stripe, line 21:
```suggestion
    const outputChannel = vscode.window.createOutputChannel('Stripe Language Client');
```

src/stripeLanguageServer/client.ts
```diff
@@ -18,22 +32,82 @@ export class StripeLanguageClient {
...
+          return ErrorAction.Continue;
+        },
+        closed: () => CloseAction.Restart,
+      },
+      revealOutputChannelOn: RevealOutputChannelOn.Error,
+    };
+
+    // Create the language client and start the client.
+    const dotnetClient = new LanguageClient(
+      'stripeCsharpLangaugeServer',
```
vcheung-stripe, line 96:
```suggestion
      'stripeCsharpLanguageServer',
```

src/stripeDotnetLanguageServer/Program.cs
```diff
@@ -0,0 +1,99 @@
...
+                                new ConfigurationItem
+                                {
+                                    Section = "typescript",
+                                }
+                            ).AddSingleton(
+                                new ConfigurationItem
+                                {
+                                    Section = "terminal",
+                                }
+                            );
```
vcheung-stripe, line 84: What are these ConfigurationItems used for?

vcheung-stripe (commented):
Makes sense to me after your explanation. The SemanticModel looks perfect for our use case. Just a couple nits and a question.

src/stripeLanguageServer/client.ts
gracegoo-stripe: Oops, me and my typos .. Updating!

src/stripeDotnetLanguageServer/Program.cs
gracegoo-stripe: Oops, leftovers from the [sample](https://github.com/OmniSharp/csharp-language-server-protocol). I think these are used to send over configs from the client -- let me remove these. Good catch!

gracegoo-stripe:
Addressed comments, and I also added another commit that fixes our intellisense issue for the subdirectory.

gracegoo-stripe:
Also a note for the build failures -- those will be fixed when I figure out how to get the dotnet executable for each platform

vcheung-stripe (approved):
Makes sense that the action fails. This LGTM.

PR #239, opened 2021-06-01
Fix tests that sometimes fail when a custom CLI path is set

vcheung-stripe:
These tests are reading the actual `stripe.cliInstallpath` on whatever system they're running on, which causes them to fail sometimes. This stubs the `getConfiguration` call so that the tests should always pass correctly.


gracegoo-stripe (approved):
nice, thanks for fixing this!

PR #240, opened 2021-06-01
Add StripeDaemon class that encapsulates the daemon logic

vcheung-stripe:
StripeDaemon handles starting the stripe daemon process, restarting it when the cli path changes, and creating the gRPC client. It exposes one method, `setupClient`, that wraps this logic for dependents of this class. The intended usage is that commands that require the stripe daemon process will use `setupClient` rather than working with the gRPC client class directly.

This code is currently unused, but will be used when Stripe Samples is implemented in a followup PR.


src/daemon/stripeDaemon.ts
```diff
@@ -0,0 +1,122 @@
...
+import {StripeClient} from '../stripeClient';
+import {Writable} from 'stream';
+import execa from 'execa';
+
+/**
+ * StripeDaemon handles starting and restarting the Stripe daemon gRPC server and creating the
+ * client connection.
+ */
+export class StripeDaemon {
+  private stripeClient: StripeClient;
```
gracegoo-stripe, line 16: We only need this for the getCLIPath currently right?  In the future, will this daemon replace `StripeClient` entirely?

gracegoo-stripe (approved):
LGTM, just one non-blocking question

src/daemon/stripeDaemon.ts
vcheung-stripe: Yep, we only need `getCLIPath` currently. The `StripeClient` definitely has tech debt and I could see `StripeDaemon` replacing it. I could also see `StripeClient` still existing, but with only `getCLIPath`, just to minimize concerns. I guess I'm ok with either but I might prefer the latter, I'm not sure.

gracegoo-stripe: Actually sorry, it's not the daemon that should replace the client, since the daemon's purpose is to manage that process -- should StripeClient actually be depending on the daemon? What would be the entry-point class that manages the requests to "stripe" -- that class should make it so that the clients (our other vscode classes) are unaware of whether we are communicating via stripe by manually kicking off cli commands via the terminal or via gRPC.

This is all non-blocking I am just thinking outloud

PR #241, opened 2021-06-02
Implement Stripe Samples creation

### LINKED ISSUE #72: As a user I can scaffold Stripe Samples from VS Code
As a user I can scaffold Stripe Samples from VS Code. 

Flow:
- "Create new project from Stripe Sample""
- User selects samples from list
- User selects their language
- Sample gets crated and opened in a new VS Code window.

vcheung-stripe:
### Summary

fixes #72 

Implement the `stripe.createStripeSample` command, available in the Quick Links view (formerly Dashboard view) and the command palette. It depends on the CLI having the `daemon` command and will prompt for an upgrade if it doesn't. It still works even when the user is logged out.

### Screenshots

Happy path. User has the daemon command and is logged in.

https://user-images.githubusercontent.com/71457708/120523429-ba187400-c38a-11eb-991d-31a0c1a4e0a0.mov

User has the daemon command but isn't logged in. We still create the sample, but we tell them to set their api keys in the .env.

https://user-images.githubusercontent.com/71457708/120523406-b2f16600-c38a-11eb-9ed5-e171d4ab04d5.mov

Prompt upgrade when the user doesn't have the daemon command

https://user-images.githubusercontent.com/71457708/120523487-cac8ea00-c38a-11eb-8989-d37966d2f6ce.mov

It's also available in the command palette

Screen Shot 2021-06-02 at 10 16 01 AM


src/extension.ts
```diff
@@ -4,14 +4,16 @@ import {ServerOptions, TransportKind} from 'vscode-languageclient';
...
-import {StripeDashboardViewProvider} from './stripeDashboardView';
+import {StripeDaemon} from './daemon/stripeDaemon';
 import {StripeDebugProvider} from './stripeDebugProvider';
 import {StripeEventTextDocumentContentProvider} from './stripeEventTextDocumentContentProvider';
 import {StripeEventsViewProvider} from './stripeEventsView';
 import {StripeHelpViewProvider} from './stripeHelpView';
 import {StripeLanguageClient} from './stripeLanguageServer/client';
 import {StripeLinter} from './stripeLinter';
 import {StripeLogsViewProvider} from './stripeLogsView';
+import {StripeQuickLinksViewProvider} from './stripeQuickLinksView';
```
gracegoo-stripe, line 15: nice, thanks for renaming the class too!

src/stripeSamples.ts
```diff
@@ -0,0 +1,302 @@
...
+    const selectedIntegrationName = await vscode.window.showQuickPick(
+      await getIntegrationNames(sample.sampleData.name),
+      {
+        placeHolder: 'Select an integration',
+      },
+    );
+
+    if (!integrations) {
+      return;
+    }
```
gracegoo-stripe, line 183: Should we return before we show the quick pick if there's no integration?

gracegoo-stripe (dismissed):
This look awesome! LGTM, with one non-blocking comment.

src/stripeSamples.ts
vcheung-stripe: That makes sense to me. Fixing.

gracegoo-stripe (approved):

Issue #220, opened 2021-04-29
API key linting false positive detects obscured keys from stripe error messages


focusaurus:
I have some error handling unit tests where I have a copy of a JSON document returned from a stripe API call. The error message includes an API key redacted with `*******`, but the linter still thinks this is a security issue.

```json
// "message": "The provided key 'sk_test_9L****************************y9yo' does not......
```

As a workaround I'll probably extra-redact this but in general I like including full sample responses in comments alongside my code. There's no actual security issue here as the key is redacted.

What I'd like to see is the linter's key matching be precise enough to ignore redacted keys.

gracegoo-stripe:
Yup good point, thanks for reporting the bug! We have addressed the issue with the linked PR.

Issue #222, opened 2021-05-06
This Stripe API Key is hardcoded. For better security, consider using a .env file.  See https://stripe.com/docs/keys#safe-keys for more advice.


dani3l3:
This message/recommendation also triggers for my actual .env file, the key is NOT in the code. 
I guess those .env files should be excluded from the scan if the recommendation is to use them... it sounds rather stupid to be scolded because you should be doing something.... that you are actually already doing.
I am using this popular library to handle the .env configuration https://github.com/vlucas/phpdotenv

gracegoo-stripe:
Thanks for reporting the bug @dani3l3! We've added logic to ignore .env files from being flagged by our linter in the PR above.

PR #245, opened 2021-06-03
Linter Updates: Ignore .env files and redacted keys 

gracegoo-stripe:
The redaction change is targeted to protect against the specific redacted format that Stripe returns which looks something like "sk_live_aa********************1234".
It does not attempt to detect other potential ways users can modify the keys  nor actually validate that the key is real.

This change fixes https://github.com/stripe/vscode-stripe/issues/220 and https://github.com/stripe/vscode-stripe/issues/222

# Testing
.env file before:
Screen Shot 2021-06-03 at 2 15 51 PM

after:
Screen Shot 2021-06-03 at 2 25 26 PM

redacted keys before:
Screen Shot 2021-06-03 at 3 24 33 PM

after:
Screen Shot 2021-06-03 at 3 11 06 PM

### DIFF
--- a/src/stripeLinter.ts
+++ b/src/stripeLinter.ts
@@ -13,8 +13,12 @@ import {Git} from './git';
 
 import {Telemetry} from './telemetry';
 
+/**
+ * Stripe returns redacted API keys in the form of sk_live_aa********************1234, so don't match unless there are
+ * at least three unredacted characters of the body of the API key.
+ */
 const stripeKeysRegex = new RegExp(
-  '(sk_test|sk_live|pk_test|pk_live|rk_test|rk_live)_[a-zA-Z0-9]+',
+  '(sk_test|sk_live|pk_test|pk_live|rk_test|rk_live)_[a-zA-Z0-9]{2}[a-zA-Z0-9]+',
   'g',
 );
 const diagnosticMessageNoGit =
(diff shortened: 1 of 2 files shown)


src/stripeLinter.ts
```diff
@@ -76,24 +79,26 @@ export class StripeLinter {
...
+        const range = new Range(index, match.index, index, match.index + match[0].length);
+        // create new diagnostic and add to the list of total diagnostics for this line of code
+        const diagnostic = new Diagnostic(range, message, severity);
+
+        this.telemetry.sendEvent('diagnostics.show', severity);
+        diagnostics.push(diagnostic);
+      }
+
+      return diagnostics;
+    };
```
gracegoo-stripe, line 103: Sorry about this, please hit "hide whitespace changes"

vcheung-stripe (approved):
Looks great! So to summarize the linter behavior for posterity:
- don't flag if the key is in a file that's git-ignored
- don't flag if the key is in a .env file
- don't flag if the key has less than 3 unredacted chars in the body

This all makes sense to me 👍

gracegoo-stripe:
> Looks great! So to summarize the linter behavior for posterity:
> 
> * don't flag if the key is in a file that's git-ignored
> * don't flag if the key is in a .env file
> * don't flag if the key has less than 3 unredacted chars in the body
> 
> This all makes sense to me 👍

Exactly!

PR #255, opened 2021-06-07
Show progress indicator while a sample's integrations are loading

### LINKED ISSUE #251: Progress Indicator while cloning stripe samples
Add a progress indicator when the cloning happens so the user gets some feedback after they've went through the samples workflow. Right now the screen just disappears and the user needs to wait for a few seconds before the workspace opening prompt pops up.

vcheung-stripe:
### Summary
fixes #251 

Tested manually.

### Screenshots
Now, the integrations prompt appears right away, and a progress indicator appears if it takes too long to load. You can see the progress indicator briefly in this video, underneath the input box.

https://user-images.githubusercontent.com/71457708/121064120-0513fc00-c77c-11eb-9582-452649c55d20.mov

### DIFF
--- a/src/stripeSamples.ts
+++ b/src/stripeSamples.ts
@@ -164,21 +164,30 @@ export class StripeSamples {
   private promptIntegration = async (
     sample: SampleQuickPickItem,
   ): Promise<SampleConfigsResponse.Integration | undefined> => {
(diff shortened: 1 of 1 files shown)


gracegoo-stripe (approved):
LGTM

Issue #253, opened 2021-06-07
Error while trying to add the stripe debugger configuration


gracegoo-stripe:
![GbuxuT5A](https://user-images.githubusercontent.com/75757829/121052022-d17ea500-c76e-11eb-8fa0-d09249900a66.png)

vcheung-stripe:
I think this can be fixed with a null check here:

https://github.com/stripe/vscode-stripe/pull/242/files#diff-ca92d1afe836a483259dfc56e55ada3e33076f195fd4f2e7a0fd06be4bfa664fR131

Change `if (options.skipVerify)` => `if (options?.skipVerify)`

PR #259, opened 2021-06-07
Adding null check for skipverify

ctrudeau-stripe:
### Summary

When you launch cmd+shift+p and run “Start webhooks event listening” the options can be null because the user will type them in. This adds a null check. 

###Testing 

Verified manually without check: 

Screen Shot 2021-06-07 at 4 22 06 PM

With check everything runs: 

Screen Shot 2021-06-07 at 4 22 59 PM

r? @vcheung-stripe 
cc @stripe/developer-products

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -128,7 +128,7 @@ export class Commands {
     }
 
     const skipVerify = await (async () => {
-      if (options.skipVerify !== undefined) {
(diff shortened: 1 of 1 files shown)


ctrudeau-stripe:
Fixes #253

vcheung-stripe (approved):

PR #267, opened 2021-06-09
Fix split terminal bug

### LINKED ISSUE #266: Bug in split terminal when you "Forward events ..." more than once
When you click "Forward events ..." the first time, a terminal is created that runs `stripe listen`. When you click it again, a new terminal is split from that one. There is a bug where the new split terminal isn't sent the `stripe listen` command.

As there's no official vscode API for splitting a terminal, we have been relying on this hack: https://github.com/microsoft/vscode/issues/45407#issuecomment-638493501

Something must have changed in vscode since we implemented that. We should fix it.

We should also move an official vscode API for splitting terminals if that lands.

vcheung-stripe:
fixes #266 

See the linked issue for an explanation.

### DIFF
--- a/src/stripeTerminal.ts
+++ b/src/stripeTerminal.ts
@@ -61,7 +61,7 @@ export class StripeTerminal {
     if (this.terminals.length > 0) {
       const terminal = await this.createNewSplitTerminal();
       if (!terminal) {
-        throw new Error('Failed to split Stripe terminal');
+        throw new Error('Failed to create a terminal for this Stripe command. Please try again.');
       }
 
       this.terminals.push(terminal);
--- a/src/stripeTerminal.ts
+++ b/src/stripeTerminal.ts
@@ -33,28 +33,37 @@ export class StripeTerminal {
 
     const commandString = [cliPath, command, ...args, ...globalCLIFLags].join(' ');
 
-    const terminal = await this.createTerminal();
-    terminal.sendText(commandString);
-    terminal.show();
+    try {
+      const terminal = await this.createTerminal();
+      terminal.sendText(commandString);
+      terminal.show();
+    } catch (e) {
+      vscode.window.showErrorMessage(e.message);
+    }
   }
 
-  private createNewSplitTerminal(): Promise<vscode.Terminal> {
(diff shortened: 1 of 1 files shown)


src/stripeTerminal.ts
```diff
@@ -33,28 +33,37 @@ export class StripeTerminal {
...
+    return vscode.window.activeTerminal;
   }
 
   private async createTerminal(): Promise<vscode.Terminal> {
     if (this.terminals.length > 0) {
-      const lastTerminal = this.terminals[this.terminals.length - 1];
-      lastTerminal.show(); // In case it is hidden
       const terminal = await this.createNewSplitTerminal();
+      if (!terminal) {
+        throw new Error('Failed to split Stripe terminal');
```
ctrudeau-stripe, line 64: Should we say try again on error? I'm slightly confused what a Stripe terminal is. What do you think?

ctrudeau-stripe (approved):
LGTM, slight change on the wording maybe?

src/stripeTerminal.ts
vcheung-stripe: Good question. I don't think we want to retry because to reach this exception, then something very wrong must have happened.

A Stripe terminal is a vscode terminal whose sole purpose is to run `stripe` commands. I'm not sure if there's a better name for it, happy to change it if you have any suggestions.

vcheung-stripe: Oops sorry, I misread your comment - yeah we can say try again on error! And then we won't have to mention anything about a stripe terminal.

src/stripeTerminal.ts
```diff
@@ -61,7 +61,7 @@ export class StripeTerminal {
     if (this.terminals.length > 0) {
       const terminal = await this.createNewSplitTerminal();
       if (!terminal) {
-        throw new Error('Failed to split Stripe terminal');
+        throw new Error('Failed to create a terminal for this Stripe command. Please try again.');
```
ctrudeau-stripe, line 64: Amazing!

ctrudeau-stripe (approved):
LGTM :shipit:

PR #268, opened 2021-06-09
Update event details JSON to resemble Dashboard

vcheung-stripe:
Due to some limitations with the gRPC server, certain fields in objects that should be displayed as `null` are instead displayed as `''`. This change converts `''` to `null` to be consistent with the Events page in the Dashboard.

Added test for the utils and manually tested the JSON display.


src/stripeEventsView.ts
```diff
@@ -140,15 +146,17 @@ export class StripeEventsViewProvider extends StreamingViewDataProvider<ListenRe
...
 
-    // Unfortunately these steps are necessary for correct rendering
+    // Unfortunately, the following steps are necessary for correct JSON formatting. We want the
+    // JSON to look like what a user sees in the Stripe Dashboard.
     const stripeEventObj = {
       ...stripeEvent.toObject(),
       data: stripeEvent.getData()?.toJavaScript(),
     };
     const snakeCaseStripeEventObj = recursivelyRenameKeys(stripeEventObj, camelToSnakeCase);
+    const finalStripeEventObj = recursivelyMapValues(snakeCaseStripeEventObj, emptyStringToNull);
```
gracegoo-stripe, line 156: Are there cases where there are genuinely empty strings on the dashboard view? Would it be possible for us to create a custom serializing function for the Data field? I assume the toJavascript() is also what's converting the snake case to camel case that we need to convert again?

vcheung-stripe: - I checked with @aidan-stripe from API services and we can't reasonably conclude there are no empty strings. So I'll have to change this.
- It turns out `toJavascript()` is not what's converting snake case to camel case. I think that has to do with the protobuf compiler. However it turns out `toJavascript()` is what's actually causing the empty string issue!
- I'll remove `toJavascript()` and I think I'll go with implementing a custom serializing function for `data`. That feels like a good approach.

vcheung-stripe: Actually I think we are fine here. `toJavascript()` is setting null values correctly on the `data` field. What is *technically* incorrect is certain fields in the metadata (e.g. `account`, `idempotency_key`) were being set with empty string when they should be `null`. However, these fields aren't shown in the dashboard so there is no inconsistency problem.

I'm closing this PR as this is a non-issue.

PR #269, opened 2021-06-09
Display subset of log details JSON in the editor

### LINKED ISSUE #247: Show log entry inside the editor
When clicking on logs today, it takes you to the page on the dashboard. What we want is to be able to click on the log entry and have it open in a new tab, like the events.

vcheung-stripe:
### Summary
fixes #247 

- Similar to events, this shows log details inside the editor. Unfortunately this does not show requests and responses, but it does show error responses if there are any.
- Each `req_` has a link to the log in the dashboard.

The log details, with link:
Screen Shot 2021-06-09 at 3 56 34 PM

The log details for an error:
Screen Shot 2021-06-09 at 3 56 48 PM

Right-clicking a log item still works:
Screen Shot 2021-06-09 at 3 57 10 PM

### Testing
Added tests for the workspace utils and manually verified the JSON display.


src/stripeLogTextDocumentContentProvider.ts
```diff
@@ -0,0 +1,39 @@
+import * as vscode from 'vscode';
```
vcheung-stripe, line 1: This file is copied from the one for events. There might be a way to share the logic with a base class or something.

vcheung-stripe:
Note that the tooltip is configurable. We could put descriptive text here.

Screen Shot 2021-06-09 at 5 17 50 PM

src/stripeLogTextDocumentContentProvider.ts
gracegoo-stripe: I think we can do this with one class too: 

1. Keep two schemas and register two instances of a `StripeResourceDocumentContentProvider`. The `StripeResourceDocumentContentProvider` takes in the regex pattern and a function to fetch the object from the extension context in its constructor.

or 

2. We have one schema, and one instance of a `StripeResourceDocumentContentProvider`. But that class can figure out on its own what type of resource the resourceId represents is and how to fetch it. Maybe it can have a mapping of the regex match to retrieve function, and it tries to loop through that mapping to see if any match the resourceId it was given.


WDTY? I think I'd vote for 1 just because 2 defers more logic to happen at runtime.

src/stripeLogsDashboardLinkProvider.ts
```diff
@@ -0,0 +1,37 @@
+import * as vscode from 'vscode';
+import {LOG_ID_REGEXP} from './resourceIDs';
+
+export class StripeLogsDashboardLinkProvider implements vscode.DocumentLinkProvider {
```
gracegoo-stripe, line 4: neat!

gracegoo-stripe (commented):
Nice, I love that you've added links inside the doc!

vcheung-stripe:
@gracegoo-stripe I implemented what you suggested. Have one class that takes a regexp and retrieve function. Instantiate two instances of it, one for logs and one for events.

vcheung-stripe:
@gracegoo-stripe I had to rebase to fix tests after merging https://github.com/stripe/vscode-stripe/commit/6b1db29edd4ca4499f7dc7b04ff1e8121571e1f3 (which I accidentally force-pushed to master, but that's a different story). Can you stamp?

gracegoo-stripe (approved):

PR #270, opened 2021-06-10
Updating the text for the webhooks prompt

### LINKED ISSUE #248: Update Forward webhooks prompting UX
When the user opens up the forward webhooks dialogue, we prompt the user to enter in two addresses, the 2nd for connect. The message varies only slightly so it's easy for users to think their input got eaten. 

We need to update the UX to at least clarify this.

ctrudeau-stripe:
Screen Shot 2021-06-09 at 5 07 24 PM

Screen Shot 2021-06-09 at 5 07 20 PM

r? @vcheung-stripe 

fixes #248 updating the text to show connect vs account more prominently. Given a lack of better UX text is all we can do unless we go full custom.

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -100,12 +100,13 @@ export class Commands {
       const defaultForwardToURL = 'http://localhost:3000';
 
       const forwardToInput = await vscode.window.showInputBox({
-        prompt: 'Enter local server URL to forward webhook events to',
+        prompt:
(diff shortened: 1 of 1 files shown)


vcheung-stripe:
Looks like you have a transient test failure. Restarting it should work.

vcheung-stripe (approved):

PR #271, opened 2021-06-10
Syncing Solution and improving some dev workflows

gracegoo-stripe:
This commit adds the ability for us to update the Solution as we get events for changes in files we are interested in. It also:
* Creates a workspace configuration so we can have separate vscode settings for the dotnet project.
* Adds a Unit Test project (in dotnet, Tests are usually in separate projects from the source code)
* Move the source code one layer deeper into another folder called `stripe.LanguageServer`

Demo:

https://user-images.githubusercontent.com/75757829/121451499-8cb36380-c952-11eb-99d6-c6b94a5d3417.mp4


gracegoo-stripe:
Builds will keep failing for the same reason until I figure out how to acquire the dotnet executable in all platforms.

vcheung-stripe (approved):
Syncing on change and on open make sense to me. Minor non-blocking nit, otherwise LGTM! 👍

src/stripeDotnetLanguageServer/stripe.LanguageServer/WorkspaceManager.cs
```diff
@@ -0,0 +1,87 @@
...
+            var documentId = this._currentSolution.GetDocumentIdsWithFilePath(uri.GetFileSystemPath()).FirstOrDefault();
+            _logger.LogDebug("Getting DocumentId: " + documentId);
+            return this._currentSolution.GetDocument(documentId);
+        }
+
+        // Handles OpenDocument events
+        // If document already exists in the project, return it. If document does not exist, it can be for two reasons:
+        //  1. the .cs file is a metdata file that was opened. i.e. a library reference. In this case, the scheme of the uri returned will be "omnisharp-metadata"
+        //  2. the .cs file is a new file created for the project.
+        public void OpenDocument(DocumentUri uri)
```
vcheung-stripe, line 48: Nit: These are already readable, but I feel it could be more readable if they were called `HandleDidOpenTextDocument` and `HandleDidUpdateTextDocument`. But I don't know anything about C# conventions so this is perfectly fine to me

PR #275, opened 2021-06-15
Adding conditional language server startup and passing in project from user workspace

gracegoo-stripe:
I also moved the language server client out of the stripeLanguageServer directory because I think it belongs in the extension code.

# Testing

https://user-images.githubusercontent.com/75757829/122124584-552a3880-cde4-11eb-8146-5b2cd3161b26.mp4

### DIFF
--- a/src/languageServerClient.ts
+++ b/src/languageServerClient.ts
@@ -5,20 +5,34 @@ import {
   ErrorAction,
   LanguageClient,
   LanguageClientOptions,
-  RevealOutputChannelOn,
   ServerOptions,
   Trace,
 } from 'vscode-languageclient';
-
-import {Telemetry} from '../telemetry';
+import {Telemetry} from './telemetry';
 
 export class StripeLanguageClient {
-  static activate(
+  static async activate(
     context: vscode.ExtensionContext,
     serverOptions: ServerOptions,
     telemetry: Telemetry,
   ) {
     const outputChannel = vscode.window.createOutputChannel('Stripe Language Client');
+
+    // start the csharp server if this is a dotnet project
+    const dotnetProjectFile = await this.getDotnetProjectFiles();
+    if (dotnetProjectFile.length > 0) {
+      this.activateDotNetServer(context, outputChannel, dotnetProjectFile[0]);
+    } else {
+      this.activateUniversalServer(context, outputChannel, serverOptions, telemetry);
+    }
+  }
+
+  static activateUniversalServer(
+    context: vscode.ExtensionContext,
+    outputChannel: vscode.OutputChannel,
+    serverOptions: ServerOptions,
+    telemetry: Telemetry,
+  ) {
     outputChannel.appendLine('Starting universal client');
     const universalClientOptions: LanguageClientOptions = {
       // Register the server for stripe-supported languages. dotnet is not yet supported.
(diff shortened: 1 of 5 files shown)


gracegoo-stripe:
Same deal with the builds -- will not pass on other platforms until we update our executable acquisition logic.

src/languageServerClient.ts
```diff
@@ -109,4 +120,41 @@ export class StripeLanguageClient {
...
     outputChannel.appendLine('C# language service is running.');
   }
+
+  /**
+   * Returns a solutions file or project file if it exists in the workspace.
+   *
+   * If the user is working with multiple workspaces that contain C# projects, we don't know which one to run the server on, so we'll return the first one we find.
+   * In the future, we may want to prompt the user to pick one, or expose a command that let's the user change the tracking project and restart.
+   *
+   * Returns null if none of the workspaces are .NET projects.
```
vcheung-stripe, line 130:
```suggestion
   * Returns [] if none of the workspaces are .NET projects.
```

test/suite/languageServerClient.test.ts
```diff
@@ -0,0 +1,107 @@
...
+  teardown(() => {
+    sandbox.restore();
+  });
+
+  suite('getDotnetProjectFiles', () => {
+    const workspacePath = '/my/path';
+
+    test('returns empty when no workspaces', async () => {
+      sandbox.stub(vscode.workspace, 'workspaceFolders').value([]);
+      console.log('BAOASPJD');
```
vcheung-stripe, line 22: Nit: remove console.log

test/suite/languageServerClient.test.ts
```diff
@@ -0,0 +1,107 @@
...
+      sandbox.stub(vscode.workspace, 'workspaceFolders').value([
+        {
+          index: 0,
+          name: 'workspace',
+          uri: vscode.Uri.file(workspacePath),
+        },
+      ]);
+      const fileFilesStub = sandbox.stub(vscode.workspace, 'findFiles');
+      fileFilesStub.onCall(0).returns(Promise.resolve([]));
+      fileFilesStub.onCall(1).returns(Promise.resolve([vscode.Uri.file(csprojFile)]));
```
vcheung-stripe, line 61: Might be worth commenting what the first and second calls correspond to. Or better, we could make `findFilesStub` return different values depending on the argument.

vcheung-stripe (approved):
Nice! Just a couple small things, but they're non-blocking.

test/suite/languageServerClient.test.ts
gracegoo-stripe: haha oops, thank you

src/languageServerClient.ts
gracegoo-stripe: Good catch!

test/suite/languageServerClient.test.ts
gracegoo-stripe: Yeah, I tried the latter at first but I could not get that stub to be invoked when I provided the arguments. Let me give it another try and then I'll just leave some comments if that doesn't work out

gracegoo-stripe:
Addressed comments, ptal @vcheung-stripe

vcheung-stripe (approved):

PR #276, opened 2021-06-16
Add publishing step for dotnet language server

gracegoo-stripe:
This commit adds a new npm script target that calls the `dotnet publish` command to place all necessary libs into the dist/stripeDotnetLanguageServer folder which we can then access from the client. We are excluding the test project from being included in the publishing. 

I also updated the task.json as well to use the npm script. 

Note -- this did not need to be an npm script at all, we could have just added a bash script that has the same contents and called it, but I figured it'll just be easier to keep track of if all our build steps were in the same place. 

# Testing 
1. Made sure the language server can startup via the debugger launcher 
2. Packaged up the extension with `vsce package`, manually installed it, and made sure the language server can start up.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -33,15 +33,9 @@
     },
     {
       "label": "build dotnet server",
-      "command": "dotnet",
-      "type": "process",
-      "args": [
-          "build",
-          "${workspaceFolder}/src/stripeDotnetLanguageServer/stripe.LanguageServer/stripe.LanguageServer.csproj",
-          "/property:GenerateFullPaths=true",
-          "/consoleloggerparameters:NoSummary"
-      ],
-      "problemMatcher": "$msCompile"
+      "type": "npm",
+      "script": "publish:dotnet-server",
+      "isBackground": true
   },
     {
       "label": "Build all",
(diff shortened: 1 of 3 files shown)


vcheung-stripe (approved):
LGTM

gracegoo-stripe:
I forgot to check the size of the extension afterwards and it's gone up from 2 to 10MB. I'm looking around to see if we can bring that down. 

There are some optimization flags available with dotnet publish, but those are only available for [self-contained](https://docs.microsoft.com/en-us/dotnet/core/deploying/) applications. These allow us to include a runtime executable within the publishing (which would fix our other problem of acquiring the dotnet execuable at runtime), but we would need to make one for every platform in that case. I'll run a test tomorrow to see how the sizes would compare.

gracegoo-stripe:
Okay, ran a test and here are the numbers:
* Without the dotnet app: 2MB
* With the app: 10MB
* Standalone app (including the dotnet runtime) for linux-x64: 40MB
* Standalone app for linux-x64 with trimming and single-file enabled: 22MB

So I'd say the self-contained option is not worth it. One is already still larger than the app and we'd need to ship with multiple self-contained apps which would increase the extension size even more. 

I'll merge this version in

PR #281, opened 2021-06-17
Using the .NET runtime extension to acquire our executable 

gracegoo-stripe:
Using dotnet's install tool to get our dotnet runtime which we will use to kick off our language server process: https://marketplace.visualstudio.com/items?itemName=ms-dotnettools.vscode-dotnet-runtime

* Removing the logging from saving to /tmp for now, I want to figure out how to pipe that into the client output directly so it shows up in the editor in a future PR. 
* I had to remove a couple of unit tests. I've verified that the tests are failing because of the way we are setting up the tests.
  * The extension activation test was failing because it can't find the dependent extension because we're manually activating just our extension: https://github.com/microsoft/vscode-extension-vscode/issues/74
  * the languageClient tests fail on windows because of the way we are setting up the workspace paths 
I'll do a fast follow-up for these two issues. 

# Testing

dotnet-server (1)


vcheung-stripe (approved):

vcheung-stripe:
Do we expect the tests to pass on this brtanch going forward?

gracegoo-stripe:
> Do we expect the tests to pass on this brtanch going forward?

Yes!

PR #283, opened 2021-06-18
Fixing Stripe Samples folder opening on Windows

gracegoo:
From the API doc: 
The *difference* between `Uri#parse` and `Uri#file` is that the latter treats the argument as path, not as stringified-uri.
Also verified that opening samples still works on mac.

Before:
https://user-images.githubusercontent.com/2405770/122610962-5bfbba00-d035-11eb-8661-352549cae4d6.mp4

After:
https://user-images.githubusercontent.com/2405770/122610964-5dc57d80-d035-11eb-8f50-d2c9c16fde5a.mp4

### DIFF
--- a/src/stripeSamples.ts
+++ b/src/stripeSamples.ts
@@ -287,12 +287,12 @@ export class StripeSamples {
 
     switch (selectedOption) {
       case openFolderOptions.sameWindow:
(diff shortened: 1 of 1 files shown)


ianjabour-stripe (approved):
LGTM!

Issue #287, opened 2021-06-21
Installing extension seems to expect .NET tools, and fails


elihorne:
First time user of the extension. I mostly work with JS (node and front end). When installing the extension, I get the following error message from VS Code (see screenshot).  Uninstalling the Stripe extension and re-installing provides the same result.

"An error occurred while installing .NET (5.0): .NET acquisition failed: Installation failed: Error: Command failed..."

<img width="467" alt="CleanShot 2021-06-21 at 16 43 35@2x" src="https://user-images.githubusercontent.com/51566/122825375-d8013680-d2af-11eb-8e06-90093877f88f.png">

Perhaps user error, but for my level of usage, requiring .NET tooling feels like heavy overhead, and I'm disinclined to continue since other extensions don't seem to have required this.

gracegoo-stripe:
Hey Eli, apologies you ran into this on your first install and thank you for reporting this. Your assumption is correct, the ideal scenario is that we only require .NET tooling if we detect that you are working on a .NET project. There are two things I'll need to look into here:

1. Why the extension tries to acquire a .NET runtime when you're working on a node project. Under the hood, we have some logic that starts up a .NET language server if we detect a .sln or .csproj file, can you verify that the project you had open did not have files with these extension in your workspace? 
2. Why the dotnet acquisition failed -- Looks like you are on macOS, but could you confirm? 

At this moment, I don't think the failure to acquire the runtime should be so loud either so I'll look into that as well.

Lastly, you should still be able to get all of the features the extension has to offer if you're developing on node, so clicking on the "Don't show again" button should let you proceed. Please let me know if that is not the case and if you run into any other issues!

elihorne:
Thanks for the fast reply!

I can confirm that the project context I was in uses Node for development and serving a variety of front end frameworks (Vue, etc) and that I'm on MacOS - Latest version, Apple M1 chip. No .NET development happening anywhere on my machine, to my knowledge.

I'll click "don't show again" and see what happens.

gracegoo-stripe:
I tested this out on a bigger repo I work on and noticed that it caught a sln file in my node_modules folder! So I will update our solution finding logic to make sure that does not happen anymore.

I tried to reproduce the .NET runtime download failure but was unable to do so. I did notice that the extension logs to its own output channel though, so if you are able to find those please let us know! I'm hoping it'll tell us where it exactly tried to download from. You can find the logs by going to View -> Output -> .NET Runtime (in the dropdown).

[...]

PR #288, opened 2021-06-21
Adding exclude pattern for dotnet project finder

gracegoo-stripe:
Addresses part of https://github.com/stripe/vscode-stripe/issues/287
We want to exclude common folders that are not part of the user's workspace when we try to detect solutions. There may be more we want to add to the exclusion list later on. 

# Testing
Before this change, this file triggering the dotnet server startup `node_modules/node-sass/src/libsass/win/libsass.sln` but after this change it does not


vcheung-stripe (approved):
I did not know node_modules would be included in the workspace. LGTM.

PR #291, opened 2021-06-23
Do not try to acquire dotnet runtime if we are on M1 macs 

gracegoo-stripe:
Fixes https://github.com/stripe/vscode-stripe/issues/287.

dotnet < 6.0 is not supported for native M1 so this commit adds a check to not even attempt to acquire it.
Given that dotnet is not supported, we don't expect this path to be hit anyway but adding telemetry to get some numbers.

* Also added some unit tests for this change and also fixed the unit tests for the `getDotnetProjectFiles` and added them back in.

### DIFF
--- a/src/languageServerClient.ts
+++ b/src/languageServerClient.ts
@@ -9,6 +9,7 @@ import {
   ServerOptions,
   Trace,
 } from 'vscode-languageclient';
+import {OSType, getOSType} from './utils';
 import {Telemetry} from './telemetry';
 
 export class StripeLanguageClient {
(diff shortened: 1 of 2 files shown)


src/languageServerClient.ts
```diff
@@ -74,8 +76,16 @@ export class StripeLanguageClient {
     telemetry: Telemetry,
   ) {
     outputChannel.appendLine('Detected C# Project file: ' + projectFile);
-
     const dotnetRuntimeVersion = '5.0';
+
+    // arm64 is not supported for dotnet < 6.0:
+    // https://github.com/dotnet/core/issues/4879#issuecomment-729046912
+    if (getOSType() === OSType.macOSarm) {
+      outputChannel.appendLine(`.NET runtime v${dotnetRuntimeVersion} is not supported for M1`);
```
tjfontaine-stripe, line 84: You might want to say `Apple Silicon`

vcheung-stripe (approved):
LGTM. Unfortunate that we don't have M1 macs to test this with.

PR #295, opened 2021-07-12
Create dependabot.yml

gracegoo-stripe:
An entry for the top level extension and another for the dotnet app. Leaving in a separate target branch to test out the scanning on this branch: https://github.com/stripe/vscode-stripe/tree/gracegoo-package-scan


vcheung-stripe (approved):

gracegoo-stripe:
I don't think there is a way to rebase from the web UI so instead of just re-opening a new PR I pushed it since we know the test failure came from another issue: https://github.com/stripe/vscode-stripe/pull/296

PR #296, opened 2021-07-13
Resolving samples information box

gracegoo-stripe:
A test recently began to fail in some platforms intermittently after 1.58.1. Upon [investigation with logs](https://github.com/stripe/vscode-stripe/actions/runs/1028044302), the failing test was never getting the quick pick item selections. Running the failing test by itself would [pass](https://github.com/stripe/vscode-stripe/actions/runs/1028125520). 

My guess is that something changed with the showInformation window that will cause it to hang if we don't pick a prompted selection. I made the window return quickly with a selected value which then allowed the subsequent test to run as it did before.

### DIFF
--- a/test/suite/stripeSamples.test.ts
+++ b/test/suite/stripeSamples.test.ts
@@ -85,8 +85,9 @@ suite('StripeSamples', function () {
         const showOpenDialogStub = sandbox
           .stub(vscode.window, 'showOpenDialog')
           .resolves([vscode.Uri.parse('/my/path')]);
-        const showInformationMessageSpy = sandbox.spy(vscode.window, 'showInformationMessage');
-
+        const showInformationMessageStub = sandbox
+          .stub(vscode.window, 'showInformationMessage')
+          .resolves();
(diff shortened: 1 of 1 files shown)


vcheung-stripe (approved):
Good catch. Another possibility I was thinking of was a race condition between `stripeSamples.selectAndCloneSample()` and `await simulateSelectAll()`. If we see this failing again later, I'd want to investigate that.

gracegoo-stripe:
> Good catch. Another possibility I was thinking of was a race condition between `stripeSamples.selectAndCloneSample()` and `await simulateSelectAll()`. If we see this failing again later, I'd want to investigate that.

Yes! I wondered this too and agree there's a chance that we don't get to that first prompt before the first sleep is over. In this case though, the first test to go through that logic would always pass which led me to thinking there was some unfinished state between the tests.

gracegoo-stripe:
It occurred to me now that maybe a better way of doing this is to open a fresh workspace in between test cases to make sure there aren't any lingering states in between. We can investigate that route too if this happens again.

PR #336, opened 2021-07-21
initializing osName once in telemetry

gracegoo-stripe:
osName() can sometimes take a while in the Windows Server 2019 OS we use for CI. We use it in `sendEvent` multiple times so this change initializes it once during construction of the telemetry object.


vcheung-stripe (approved):
LGTM

PR #351, opened 2021-08-19
Updating various packages

gracegoo-stripe:
Updating a bunch of the npm packages. The rest are upgrades that will be a bit more involved so I'll tackle those separately.


vcheung-stripe (approved):

gracegoo-stripe:
Oops, the upgrade command did not do what I thought so I'll be having another PR out shortly.

Issue #217, opened 2021-04-21
Keep supported triggers in sync with the CLI 


gracegoo-stripe:
Right now the supported list of events lives as a static list in our extension: https://github.com/stripe/vscode-stripe/blob/e2f83891311cefacc6408d527cfafb20a77773dd/src/commands.ts#L26

We need a better way to keep this in sync. Couple of options:
1. Make a call to the CLI to grab the list of supported events at extension startup. The help documentation for `stripe trigger` does spit out the list of events that we can parse, or we can add a special subcommand to triggers to give us this list. 
2. Create a build target to update this list.

PR #368, opened 2021-09-14
Dynamically retrieve supported triggers list from CLI

etsai-stripe:
## Summary
context: https://github.com/stripe/vscode-stripe/issues/217

Supported list of events was a static list in our extension:
`vscode-stripe/src/commands.ts`
```
 supportedEvents: string[] = [ 
We need a better way to keep this in sync. Couple of options:
```

To dynamically retrieve the supported triggered events list, make a call to the CLI gPRC service at trigger event click. if the list is failed to be retrieved, then fall back to the old static list and outputs an warning to output window

Screen Shot 2021-09-14 at 1 57 24 PM
Screen Shot 2021-09-14 at 11 44 38 AM

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -19,7 +19,7 @@ import {StripeTerminal} from './stripeTerminal';
 import {StripeTreeItem} from './stripeTreeItem';
 import {SurveyPrompt} from './surveyPrompt';
 import {Telemetry} from './telemetry';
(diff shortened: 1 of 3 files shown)


gracegoo-stripe (dismissed):
This looks perfect!

gracegoo-stripe (approved):
LGTM, thanks for fixing the tests!

PR #376, opened 2021-10-01
DX-6801: Language API Refs

etsai-stripe:
Generate new language api ref for java and update existing dotnet/universal ones


config/api_ref/patterns_java.json
```diff
@@ -0,0 +1,157 @@
+{
+  "BalanceTransaction\\.retrieve\\(": "/balance_transactions/retrieve",
```
gracegoo-stripe, line 2: Will these be updated later? These patterns won't work for cases where the Stripe method is invoked on an instance method like `charge.update`: https://stripe.com/docs/api/charges/update?lang=java
And with the language server interpreting the type of the instance the method is invoked on, will they still need to be regex patterns?

gracegoo-stripe (dismissed):
LGTM!

gracegoo-stripe (approved):

PR #377, opened 2021-10-05
DX-6802: download and extract jdt server

etsai-stripe:
Screen Shot 2021-10-04 at 6 06 01 PM
Screen Shot 2021-10-04 at 6 43 52 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -421,7 +421,8 @@
     "webpack-prod:extension": "webpack --mode production --config ./webpack.config.js && webpack --mode production --config ./src/stripeLanguageServer/webpack.config.js",
     "webpack-prod:language-server": "webpack --mode production --config ./src/stripeLanguageServer/webpack.config.js",
     "webpack-prod": "npm run webpack-prod:extension && npm run webpack-prod:language-server",
-    "publish:dotnet-server": "dotnet publish src/stripeDotnetLanguageServer/stripe.LanguageServer/ -o ./dist/stripeDotnetLanguageServer"
+    "publish:dotnet-server": "dotnet publish src/stripeDotnetLanguageServer/stripe.LanguageServer/ -o ./dist/stripeDotnetLanguageServer",
+    "java-server": "node ./out/src/stripeJavaLanguageServer/extractServer.js"
   },
   "devDependencies": {
     "@types/byline": "^4.2.33",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -461,13 +462,17 @@
...
     "vscode-languageclient": "^6.1.3",
     "vscode-languageserver": "^6.1.1",
-    "vscode-languageserver-textdocument": "^1.0.1"
+    "vscode-languageserver-textdocument": "^1.0.1",
+    "zlib": "*"
   },
   "extensionDependencies": [
-    "ms-dotnettools.vscode-dotnet-runtime"
+    "ms-dotnettools.vscode-dotnet-runtime",
+    "yaozheng.vscode-pde"
```
gracegoo-stripe, line 476: Is this extension used in the extracting script?

src/stripeJavaLanguageServer/extractServer.ts
```diff
@@ -0,0 +1,55 @@
...
+
+// Eclipse JDT source
+// https://www.eclipse.org/community/eclipse_newsletter/2017/may/article4.php
+const href = 'http://download.eclipse.org/jdtls/snapshots/';
+const tarFile = 'jdt-language-server-latest.tar.gz';
+
+const tarPath = `./${tarFile}`;
+const source = `${href}/${tarFile}`;
+
+const extractTo = './out/src/stripeJavaLanguageServer';
```
gracegoo-stripe, line 19: When we publish the extension, it'll ignore the contents of the out folder: https://github.com/stripe/vscode-stripe/blob/master/.vscodeignore

I think that makes sense. Since we are still prototyping, we don't need to publish the java language server stuff out when we make releases. But just noting here that we may want to also extract this out to the `dist` folder when we are ready to release

package.json
etsai-stripe: its a pre-requisite to run the jdt server

src/stripeJavaLanguageServer/extractServer.ts
etsai-stripe: https://jira.corp.stripe.com/browse/DX-6804

package.json
gracegoo-stripe: Ahh gotcha! When we added the ms-dotnettools.vscode-dotnet-runtime we had to go through a security review, should we wait for that before committing this to master? Releasing this as is means that all users who get the update would also automatically download this extension which sounds like is not totally necessary right now.

etsai-stripe: let me remove this for this PR then. i can add it back later when we pass review

gracegoo-stripe: We can still keep this in the PR and just not diff this to against the master branch -- I think it's still useful for progress on the java server

PR #383, opened 2021-10-11
Java Lang Server Prototype

etsai-stripe:
### Motivations
Build java lang server prototype to establish java api hover support. this is only a prototyping PR. will have a followup PR to add tests and clean up.
https://jira.corp.stripe.com/browse/DX-6795

### Prototype run recording:
https://user-images.githubusercontent.com/88805634/137003067-0d45f7d5-f0a3-433f-953a-f2553c41765c.mp4

### Hover screenshot
Screen Shot 2021-10-12 at 11 15 21 AM

### Stripe Language Client Output
Screen Shot 2021-10-12 at 1 16 09 PM


.vscode/launch.json
```diff
@@ -23,7 +23,8 @@
       "outFiles": ["${workspaceFolder}/dist/**/*.js"],
       "preLaunchTask": "Build all",
       "env": {
-        "EXTENSION_MODE": "development"
+        "EXTENSION_MODE": "development",
+        "DEBUG_VSCODE_JAVA":"true"
```
gracegoo-stripe, line 27: What does the DEBUG_VSCODE_JAVA do?

src/stripeJavaLanguageClient/utils.ts
```diff
@@ -0,0 +1,665 @@
...
+    getJavaConfiguration().get<boolean>('import.maven.enabled') || false;
+  const isGradleEnabled: boolean =
+    getJavaConfiguration().get<boolean>('import.gradle.enabled') || false;
+  if (isMavenEnabled && isGradleEnabled) {
+    const activeBuildTool: string | undefined = context.workspaceState.get(ACTIVE_BUILD_TOOL_STATE);
+    if (!activeBuildTool) {
+      if (!(await hasBuildToolConflicts())) {
+        return true;
+      }
+      outputChannel.appendLine('Build tool conflicts are detected in workspace.');
```
gracegoo-stripe, line 105: What should the users do in this case?

src/stripeJavaLanguageClient/utils.ts
```diff
@@ -0,0 +1,665 @@
...
+  });
+}
+
+async function getBuildFilesInWorkspace(): Promise<Uri[]> {
+  const buildFiles: Uri[] = [];
+  const inclusionFilePatterns: string[] = getBuildFilePatterns();
+  inclusionFilePatterns.push('**/.project');
+  const inclusionFolderPatterns: string[] = getInclusionPatternsFromNegatedExclusion();
+  // Since VS Code API does not support put negated exclusion pattern in findFiles(),
+  // here we first parse the negated exclusion to inclusion and do the search.
```
gracegoo-stripe, line 145: When is this used? Is this for cases where we want to exclude everything in `/path/a` unless it's `/path/a/b`?

etsai-stripe: this it to handle the java configuration settings. ie:
```
"java.import.exclusions": [
     "**/node_modules/**",
    "**/.metadata/**",
    "**/archetype-resources/**",
    "**/META-INF/maven/**",
    "**/Frontend/**",
],
```
if user puts `!` in front of any exclusion, it means `include`

src/stripeJavaLanguageClient/utils.ts
etsai-stripe: i will update the output to include user next step

PR #386, opened 2021-10-15
DX-6839: refactor and add tests

etsai-stripe:
https://jira.corp.stripe.com/browse/DX-6839

clean up prototype code and add tests

Java Server connection error/exception handled and shown in client IDE
Screen Shot 2021-10-19 at 5 03 41 PM

### DIFF
--- a/src/stripeJavaLanguageClient/javaServerStarter.ts
+++ b/src/stripeJavaLanguageClient/javaServerStarter.ts
@@ -6,6 +6,7 @@
 import * as fs from 'fs';
 import * as path from 'path';
 import {Executable, ExecutableOptions} from 'vscode-languageclient';
+import {ExtensionContext, OutputChannel} from 'vscode';
 import {
   JDKInfo,
   checkPathExists,
@@ -17,7 +18,7 @@ import {
   startedFromSources,
   startedInDebugMode,
 } from './utils';
-import {ExtensionContext} from 'vscode';
+import {Telemetry} from '../telemetry';
 
 export const javaServerPath = '../out/src/stripeJavaLanguageServer'; // TODO: need to move to dist folders at release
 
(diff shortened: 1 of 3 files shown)


src/stripeJavaLanguageClient/javaServerStarter.ts
```diff
@@ -76,23 +78,21 @@ function prepareParams(
...
-  });
+  const launchersFound: Array<string> = getServerLauncher(serverHome);
 
-  if (launchersFound.length) {
+  if (launchersFound && launchersFound.length) {
     params.push('-jar');
     params.push(path.resolve(serverHome, launchersFound[0]));
   } else {
-    return [];
+   return 'Server jar not found';
```
gracegoo-stripe, line 95: Instead of returning a string what if we throw an exception here? That may make it more clear that something we didn't expect happened and the caller does not have to do a type check. WDYT?

src/stripeJavaLanguageClient/javaServerStarter.ts
```diff
@@ -110,31 +111,42 @@ function prepareParams(
...
-    version = '0.0.0';
+  let version = '0.0.0';
+  try {
+    const extensionPath = path.resolve(context.extensionPath, 'package.json');
+    const packageFile = JSON.parse(fs.readFileSync(extensionPath, 'utf8'));
+    if (packageFile) {
+      version = packageFile.version;
+    }
+  } catch {
+    console.log('Cannot locate package.json');
```
gracegoo-stripe, line 138: same question re:output channel?

src/stripeJavaLanguageClient/syntaxLanguageClient.ts
```diff
@@ -31,46 +29,31 @@ export class SyntaxLanguageClient {
       },
     });
 
-    const lsPort = process.env[SYNTAXLS_CLIENT_PORT];
-    if (!serverOptions && lsPort) {
```
gracegoo-stripe, line 35: If we get rid of this, does that mean we could be starting up multiple java language servers?

src/stripeJavaLanguageClient/javaServerStarter.ts
etsai-stripe: Are you suggesting I throw a user defined exception with error message? that sounds much cleaner and i will update it!

src/stripeJavaLanguageClient/javaServerStarter.ts
etsai-stripe: same reason to why i moved out the other one. I dont have a preference either :). i will move it to output channel + telemetry

src/stripeJavaLanguageClient/syntaxLanguageClient.ts
etsai-stripe: actually yes if its in hybrid mode, then both standard and syntax servers could be started

PR #392, opened 2021-10-20
Publish Java JDT server at extension publish

etsai-stripe:


gracegoo-stripe (approved):
Awesome! How big is the extension now that we removed the tar file?

PR #414, opened 2021-11-04
Migrate login flow to use gRPC 

gracegoo-stripe:
This change updates the login flow to use the gPRC protocol and integrates the flow into the UI. 

https://user-images.githubusercontent.com/75757829/140431176-73c58014-cf37-4c80-9ab4-db07318cc990.mov

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -248,23 +248,6 @@ export class Commands {
     this.telemetry.sendEvent('login');
     const daemonClient = await stripeDaemon.setupClient();
 
-    const blah = (progress: vscode.Progress<any>) => {
-      return new Promise<void>((resolve, reject) => {
-        daemonClient.loginStatus(new LoginStatusRequest(), async (error, response) => {
-          if (error) {
-            vscode.window.showErrorMessage(`Failed to login. ${error.details}`);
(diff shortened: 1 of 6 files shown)


src/commands.ts
```diff
@@ -240,9 +244,52 @@ export class Commands {
...
+
+    daemonClient.login(new LoginRequest(), async (error, response) => {
+      if (error) {
+        vscode.window.showErrorMessage(`Failed to login. ${error.details}`);
+      } else if (response) {
+        const pairingCode = response.getPairingCode();
+        await vscode.window
+          .showInformationMessage(
+            `Your pairing code is ${pairingCode}. \n This pairing code verifies your authentication with Stripe.`,
+            ...['Authenticate from Dashboard'],
```
etsai-stripe, line 259: should we have an option to `cancel`?

etsai-stripe (approved):
LGTM! just one non blocking comment

src/commands.ts
gracegoo-stripe: Sounds good to me, adding!

etsai-stripe (approved):
thanks for making the change!

PR #424, opened 2021-11-09
DX6964: store java home path in user settings

etsai-stripe:
https://jira.corp.stripe.com/browse/DX-6964

better user experience when JDK version is missing

### updated warning message
Screen Shot 2021-11-08 at 4 51 57 PM

### stripe.java.home settings for easier jdk setup
Screen Shot 2021-11-08 at 4 38 55 PM

### prompt user to set jdk path in user settings
Screen Shot 2021-11-08 at 11 45 08 AM

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,6 +89,10 @@
           "default": true,
           "markdownDescription": "Specifies whether to enable Stripe telemetry (even if enabled still abides by the overall `#telemetry.enableTelemetry#` setting)",
           "scope": "window"
+        },
+        "stripe.java.home": {
+          "type": "string", 
+          "description": "Java home path of the JDK for Java language support"
         }
       }
     },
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -89,6 +89,10 @@
           "default": true,
           "markdownDescription": "Specifies whether to enable Stripe telemetry (even if enabled still abides by the overall `#telemetry.enableTelemetry#` setting)",
           "scope": "window"
+        },
+        "stripe.java.home": {
+          "type": "string", 
+          "description": "Java home path of the JDK for Java language support"
```
gracegoo-stripe, line 95: can we scope this to "machine" like the CLI path? This ensures that the setting can't be just set by a repo.

etsai-stripe: sure!

gracegoo-stripe (approved):

PR #445, opened 2021-11-17
dx-6943: allow user defined sample name

etsai-stripe:
https://jira.corp.stripe.com/browse/DX-6943

allow user to define the name to download the sample as. use original sample name by default.

## prompt user for name. default is sample name
Screen Shot 2021-11-17 at 9 41 59 AM

## select folder to clone. folder already has sample with the default name
Screen Shot 2021-11-17 at 9 42 24 AM

## sample clone successfully message copy updated
Screen Shot 2021-11-17 at 9 43 21 AM

## resulting folder structure
Screen Shot 2021-11-17 at 9 44 28 AM

### DIFF
--- a/src/stripeSamples.ts
+++ b/src/stripeSamples.ts
@@ -69,7 +69,9 @@ export class StripeSamples {
         return;
       }
 
-      const clonePath = await this.promptPath(selectedSample);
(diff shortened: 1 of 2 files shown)


gracegoo-stripe (approved):
look awesome!

PR #454, opened 2021-11-19
fix master CI build failures

etsai-stripe:
fix master CI build failures

use information message to show cloning in progress because the progress bar doesnt show when it is downloading too fast
Screen Shot 2021-11-18 at 7 24 46 PM

make triggering customized fixture a modal to be more obvious to user that they need to make a selection
Screen Shot 2021-11-18 at 7 25 23 PM

### DIFF
--- a/test/suite/stripeSamples.test.ts
+++ b/test/suite/stripeSamples.test.ts
@@ -15,8 +15,6 @@ import {sleep} from './helpers';
 suite('StripeSamples', function () {
   this.timeout(20000);
 
-  let sandbox: sinon.SinonSandbox;
-
   const stripeClient = <Partial<StripeClient>>{
     getCLIPath: () => Promise.resolve('/path/to/cli'),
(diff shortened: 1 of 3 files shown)


test/suite/stripeSamples.test.ts
```diff
@@ -208,7 +201,9 @@ async function simulateSelectAll() {
   await sleep(500);
   await vscode.commands.executeCommand('workbench.action.acceptSelectedQuickOpenItem');
 
+  // Pause for sample save-as name input box
+
   // Simulate select path
-  await sleep(500);
+  await sleep(1000);
```
etsai-stripe, line 207: @gracegoo-stripe 
this is the main reason for the build failure
it has passed 5 times since this fix

gracegoo-stripe (approved):
LGTM! Thanks for fixing the intermittent test failures

PR #469, opened 2021-11-30
list all webhook endpoints

etsai-stripe:
https://jira.corp.stripe.com/browse/DX-6996

- update grpc service methods
- list all webhook endpoints

Screen Shot 2021-11-30 at 11 06 24 AM


test/suite/commands.test.ts
```diff
@@ -425,8 +425,8 @@ suite('commands', function () {
     test('surfaces error from server', async () => {
       const windowSpy = sandbox.spy(vscode.window, 'showErrorMessage');
 
-      const erroMessage = 'Something went wrong';
-      const error = <grpc.ServiceError>{details: erroMessage};
+      const errorMessage = 'Something went wrong';
```
gracegoo-stripe, line 428: nice catch!

src/stripeWebhooksView.ts
```diff
@@ -0,0 +1,86 @@
...
+    //   iconPath: new ThemeIcon('add'),
+    //   tooltip: 'Create a new webhook endpoint',
+    //   contextValue: 'createWebhookEndpoint',
+    // });
+    // items.push(createEndpoint);
+
+    if (this.endpointItems.length > 0) {
+      const endpointsRootItem = new StripeTreeItem('All webhook endpoints');
+      endpointsRootItem.children = this.endpointItems;
+      endpointsRootItem.expand();
```
gracegoo-stripe, line 33: Does this mean this will start out expanded? I'm worried that it'll be too overwhelming for cases where the endpoints have all of the events enabled. Could we start this off as collapsed?

src/stripeWebhooksView.ts
```diff
@@ -0,0 +1,86 @@
...
+          // https://grpc.github.io/grpc/core/md_doc_statuscodes.html
+          // 12: UNIMPLEMENTED
+          vscode.window.showErrorMessage(
+            'Please upgrade your Stripe CLI to the latest version to use this feature.',
+          );
+        } else {
+          vscode.window.showErrorMessage(`Failed to get all webhookendpoints. ${error.details}`);
+        }
+      } else if (response) {
+        const endpoints = response.getEndpointsList();
```
gracegoo-stripe, line 55: Should we have a max # of endpoints we show? What if some merchants have a bunch of old endpoints on their test account?

src/stripeWebhooksView.ts
```diff
@@ -0,0 +1,86 @@
...
+        if (endpoints.length > 0) {
+          const endpointItems = endpoints.map((e) => {
+            const enabledEventsRootItem = new StripeTreeItem('Enabled events');
+            const enabledEvents = e.getEnabledeventsList();
+            enabledEventsRootItem.children = enabledEvents.map(
+              (event) => new StripeTreeItem(event),
+            );
+            enabledEventsRootItem.expand();
```
gracegoo-stripe, line 63: Same concern here as with the enabled events -- should we start this off as collapsed?

gracegoo-stripe (commented):
This looks amazing! Some minor comments around what we actually present in the tree

src/stripeWebhooksView.ts
etsai-stripe: how about i only show the enabled ones?

gracegoo-stripe (approved):
Looks awesome!

PR #471, opened 2021-12-03
dx6994: use new csat form

etsai-stripe:
use docs csat form to pipe data to hubble for analysis

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -1,4 +1,3 @@
-import * as querystring from 'querystring';
 import * as vscode from 'vscode';
 import {
   camelToSnakeCase,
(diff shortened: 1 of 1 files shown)


src/commands.ts
```diff
@@ -565,16 +563,7 @@ export class Commands {
...
     this.telemetry.sendEvent('openSurvey');
-    const extensionInfo = getExtensionInfo();
-
-    const query = querystring.stringify({
-      platform: encodeURIComponent(osName()),
-      vscodeVersion: encodeURIComponent(vscode.version),
-      extensionVersion: encodeURIComponent(extensionInfo.version),
-      machineId: encodeURIComponent(vscode.env.machineId),
-    });
-
```
gracegoo-stripe, line 576: Possible improvements for future iterations of the form :
1. Would it be possible to add these as questions as well and pre-fill the csat form with this info if they came from here? 
2. Similarly, could we also pre-select the VSCode tool for them by passing it in as query param?

gracegoo-stripe (approved):
Looks awesome, just a couple of suggestions for later

src/commands.ts
etsai-stripe: i will create a followup ticket to address these

PR #477, opened 2021-12-07
use generic CSAT form and pass in client data

etsai-stripe:
Screen Shot 2021-12-07 at 9 46 53 AM
Screen Shot 2021-12-07 at 9 48 15 AM


gracegoo-stripe (approved):
LGTM thanks for adding this!

PR #499, opened 2022-01-06
Retrieve and refresh integration insight on demand

etsai-stripe:
https://jira.corp.stripe.com/browse/DX-7061

dependent on PR https://github.com/stripe/stripe-cli/pull/803

### Integration insight retrieved on demand. Refreshed when log becomes available on ES
Screen Shot 2022-01-05 at 5 41 39 PM
Screen Shot 2022-01-05 at 5 42 14 PM

### Integration insight retrieved successfully. Matches the insight from dashboard
Screen Shot 2022-01-06 at 8 51 16 AM
Screen Shot 2022-01-06 at 8 51 33 AM


src/stripeWorkspaceState.ts
```diff
@@ -89,9 +91,48 @@ export function addLogDetails(
...
+async function getIntegrationInsight(logId: string, daemonClient: StripeCLIClient): Promise<string> {
+  const request = new IntegrationInsightRequest();
+  request.setLog(logId);
+
+  const integrationInsight = await new Promise<string>((resolve, reject) => {
+    daemonClient.integrationInsight(request, (error: any, response: any) => {
+      if (error) {
+        if (error.code === 12) {
+          // https://grpc.github.io/grpc/core/md_doc_statuscodes.html
+          // 12: UNIMPLEMENTED
```
vcheung-stripe, line 117: Nit: the grpc library should have these error codes built in as constants.

src/stripeWorkspaceState.ts
```diff
@@ -89,9 +91,48 @@ export function addLogDetails(
...
+          vscode.window.showErrorMessage(
+            'Please upgrade your Stripe CLI to the latest version to retrieve integration insight.',
+          );
```
vcheung-stripe, line 120: It looks like we're not resolving the promise in this case. Would this cause the log to load indefinitely?

ianjabour-stripe (approved):
LGTM - walked through changes with @etsai-stripe and everything makes sense 👍

src/stripeWorkspaceState.ts
```diff
@@ -89,9 +91,48 @@ export function addLogDetails(
...
-export function retrieveLogDetails(extensionContext: vscode.ExtensionContext, logId: string) {
+export async function retrieveLogDetails(extensionContext: vscode.ExtensionContext, logId: string, daemonClient: StripeCLIClient) {
   const logDetailsMap = getLogDetailsMap(extensionContext);
-  return logDetailsMap.get(logId);
+  const logDetails = logDetailsMap.get(logId);
+
+  // if insight has not been retrieved or previously failed to be retrieved, then retrieve it
+  if (!('insight' in logDetails) || logDetails.insight.includes('Failed to retrieve insight') || logDetails.insight.includes('Please check back later')) {
+    const insight = await getIntegrationInsight(logId, daemonClient);
+    logDetails.insight = insight;
```
vcheung-stripe, line 101: Nit: I wonder if we should avoid setting this field if there's no insight. So the user won't see an empty string. What do you think?

vcheung-stripe (commented):
I had a couple questions but it looks awesome!

src/stripeWorkspaceState.ts
etsai-stripe: currently there could be empty insight. i am trying to distinguish between `insight not retrieved yet` and `actual empty insight`, so we dont keep retrieving insight for the same log.

src/stripeWorkspaceState.ts
etsai-stripe: good catch!

src/stripeWorkspaceState.ts
etsai-stripe: i will use the grpc constants

vcheung-stripe (approved):

PR #577, opened 2022-04-15
Run build action on push only

vcheung-stripe:
I noticed that we sometimes run the build action twice for no good reason. This can slow down the time it takes to get PRs merged, especially when there are jobs waiting in queues. This changes the action to only run on [push] instead of [push, pull_request]. The idea is that in order to make a pull request, you have to push anyway, so it'll run regardless.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,6 @@
 name: Build
 
-on: [push, pull_request]
+on: [push]
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


gracegoo-stripe (approved):
That makes sense to me, thanks for making the change

vcheung-stripe:
I tried several things and could not get the status to be reported:
- changing the branch rules for master to require this action
- rerunning the action from the GitHub UI
- retriggering the action by force pushing an empty commit
- manually reporting the status via GitHub API: https://stackoverflow.com/questions/67162488/github-says-waiting-for-status-to-be-reported-for-a-valid-existing-job

I'm going to close this as it's just a nice to have.

vcheung-stripe:
Screen Shot 2022-04-15 at 1 32 32 PM

PR #823, opened 2024-02-26
Upgrade extension to latest testing libraries, update README, and VSCode version

matv-stripe:
This updates the README with instructions to how to get the extension to run.

It also updates the minimum VSCode version to allow us to use newer APIs.

It also upgrades the tests to pass in Github Actions. This required some debugging, but it ended up being because I needed to upgrade `vscode-test` to a newer version and update the platform for windows. In the future, as we update the extension, the first spot we should look should be outdated NPM packages.

I did not bump the package version just yet because I do not want to make a release (no new functionality has been added).

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "url": "https://github.com/stripe/vscode-stripe/issues"
   },
   "engines": {
-    "vscode": "^1.63.1"
+    "vscode": "^1.86.1"
   },
   "icon": "resources/logo.png",
   "categories": [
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -34,33 +34,7 @@
...
-    "onCommand:stripe.openReportIssue",
-    "onCommand:stripe.openDocs",
-    "onCommand:stripe.openSurvey",
-    "onCommand:stripe.openSamples",
-    "onCommand:stripe.startLogsStreaming",
-    "onCommand:stripe.stopLogsStreaming",
-    "onCommand:stripe.clearRecentLogs",
-    "onCommand:stripe.clearRecentEvents",
-    "onCommand:stripe.startEventsStreaming",
-    "onCommand:stripe.stopEventsStreaming",
```
vcheung-stripe, line 63: Do we need to get rid of these activation events? I suppose the `onView` ones are noop because the extension would've already been active when you view them but the `onCommand` ones seem important.

varughese: (This is @matv) - in newer version of VSCode, the linter suggested I remove these.

https://code.visualstudio.com/api/references/activation-events#onCommand

> Note: Beginning with VS Code 1.74.0, commands contributed by your extension do not require a corresponding onCommand activation event declaration for your extension to be activated.

vcheung-stripe (approved):

PR #833, opened 2024-10-09
Escape strings to be passed in as arguments.

kwu-stripe:
Addresses the bug where command injection can be used in the projectName value in workspace settings, by escaping strings passed in as arguments. 

https://security-insights.corp.stripe.com/pathfinder/sats/SAT-10878

The solution is based on [shell-escape](https://www.npmjs.com/package/shell-escape), but also adapted to handle Windows computers. This was tested on both Windows and Mac.

Screenshot 2024-10-09 at 12 36 11 PM

### DIFF
--- a/test/suite/shellEscape.test.ts
+++ b/test/suite/shellEscape.test.ts
@@ -1,3 +1,4 @@
+/* eslint-disable quotes */
 import * as assert from 'assert';
 import * as shellEscape from '../../src/shellEscape';
 import * as sinon from 'sinon';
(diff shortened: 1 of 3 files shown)


kwu-stripe:
r?

test/suite/shellEscape.test.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  setup(() => {
+    sandbox = sinon.createSandbox();
+  });
+
+  teardown(() => {
+    sandbox.restore();
+  });
+
+  suite('shellEscape', () => {
```
vcheung-stripe, line 18: To be thorough, can you also include tests for strings with quotes inside of them?

vcheung-stripe (changes requested):

vcheung-stripe (approved):

PR #836, opened 2024-11-12
Use task to run listen command

vcheung-stripe:
This uses a vscode task to run the `stripe listen` command, allowing us to leverage the built-in escaping features for the shell input: https://code.visualstudio.com/api/references/vscode-api#ShellQuotedString


charliecruzan-stripe (changes requested):
nice! i think this means we can delete the `shellEscape.ts` and `shellEscape.test.ts` files?

vcheung-stripe:
Yeah I think so, will do

charliecruzan-stripe (approved):

charliecruzan-stripe (approved):

bernardoamc-stripe:
This seems in line with what I was thinking. 👏 Do we have a way to test this with the researcher input?

vcheung-stripe:
@bernardoamc-stripe I've tried it out with the examples we've seen and they work insofar as they get quoted as expected.

PR #839, opened 2024-12-03
Escape characters in project name

vcheung-stripe:
This escapes more characters in the project name configuration value so that we keep the input correct and useful.


jdt-language-server-latest.tar.gz
matv-stripe, line 1: Can we add this to the gitignore

bernardoamc-stripe (approved):

